PR #31, opened 2018-04-27
Binary data

thewtex:


itkwidgets/trait_types.py
```diff
@@ -183,7 +182,7 @@ def itkimage_from_json(js, manager=None):
     else:
         ImageType, dtype = _type_to_image(js['imageType'])
         decompressor = zstd.ZstdDecompressor()
-        pixelBufferArrayCompressed = np.frombuffer(base64.b64decode(js['compressedBase64Data']), dtype=dtype)
+        pixelBufferArrayCompressed = np.array(js['compressedData'], dtype=dtype)
```
vidartf: Maybe you should still use `np.frombuffer`? I think that way you avoid copying the memory? This is just off the top of my head, so you should probably double check though.

thewtex: `np.frombuffer` throws an error about missing `__buffer__`.

Any other suggestions welcome...

vidartf: Is this with python 2? I'm using this code somewhere else:
```python
if six.PY2:
    # Numpy does not support frombuffer with memoryview
    # on Python 2:
    buffer = value['buffer'].tobytes()
    n = np.frombuffer(buffer, dtype=value['dtype'])
else:
    n = np.frombuffer(value['buffer'], dtype=value['dtype'])
```

vidartf: (PS: I updated the code in the previous comment to use the same variable names as your code.)

thewtex: Excellent, thanks for the suggestion, @vidartf ! Adding in a follow-up commit...

PR #36, opened 2018-05-18
ENH: ImageJ ImgLib2 example

thewtex:
Example from:

  https://github.com/CellProfiler/notebooks/blob/95a2558060075b0c8c9ad254413fac8dacfc2d12/imagej_pyjnius_demo.ipynb

Co-authored-by: Curtis Rueden <ctrueden@wisc.edu>

### DIFF
--- a/examples/ImageJImgLib2.ipynb
+++ b/examples/ImageJImgLib2.ipynb
@@ -1,0 +1,185 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Install dependencies for this example\n",
+    "#\n",
+    "# Here we assume itkwidgets has already been installed\n",
+    "import sys\n",
+    "#  todo add installation instructions for the new imagej.py"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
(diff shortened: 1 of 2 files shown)


thewtex:
CC: @ctrueden

examples/ImageJImgLib2.ipynb
```diff
@@ -31,24 +33,17 @@
...
+   "outputs": [],
    "source": [
     "# Initialize imagej. Point to a local Fiji installation\n",
+    "# Replace the value below with to your local Fiji installation.\n",
     "fiji_path = '/home/matt/bin/Fiji.app/'\n",
     "\n",
     "# Note: order is important!\n",
     "import imagej\n",
-    "imagej.quiet_init(fiji_path)\n",
+    "imagej.init(fiji_path)\n",
```
ctrueden: In addition to changing the method to `init`, I also made it create an instance of `net.imagej.ImageJ` internally, so that you can simply write:
```
ij = init(fiji_path)
```
Without all the `import imglyb` etc. rigamarole. (Remember how the order of those imports was super important to get right, or else things didn't work? Shouldn't be an issue anymore.)

Now you can just import `imglyb` the first time you need one of the conversion methods. And only import `autoclass` if/when you need to wrap additional Java classes.

thewtex: That's fantastic!

Here's a follow-up:

  https://github.com/InsightSoftwareConsortium/itk-jupyter-widgets/pull/38

PR #38, opened 2018-05-21
DOC: Simplify ImageJ initialization

thewtex:
@ctrueden please take a look.


ctrueden:
Thanks @thewtex. Code looks right to me. As long as the notebook still runs, then :+1: to merge.

PR #39, opened 2018-05-21
DOC: Improve ImageJ installation grammar

thewtex:


examples/ImageJImgLib2.ipynb
```diff
@@ -37,8 +37,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# Initialize imagej. Point to a local Fiji installation\n",
-    "# Replace the value below with to your local Fiji installation.\n",
+    "# Initialize imagej.\n",
+    "# Replace the value below to your local Fiji installation.\n",
```
jcfr: what about  `Replace the value below with the path to your local Fiji installation`

thewtex: @jcfr yes, that is right -- thanks!

Issue #44, opened 2018-05-25
Add dask array support


thewtex:
Stages:

1. [ ] Basic support for `dask.array`
2. [ ] Stream `dask.array`
3. [ ] Use `dask.delayed` or `dask.futures` to help downsample images.

CC: @mrocklin @dani-lbnl
xref #43 https://github.com/scisprints/2018_05_sklearn_skimage_dask/issues/12

thewtex:
@mrocklin good suggestion regarding `coarsen`! :+1: 

Some experiments are summarized here:

https://gist.github.com/thewtex/079c59ea71a1da59bea5e14cb6b73d1f

For large image support it looks like we should not convert NumPy arrays, ITK Images, etc. to Dask for downsampling because of the extra memory required, the time required for chunking, and the performance of downsampling. However, I will look into adding first class Dask array support by using the Dask arrays directly for large images and `coarsen`. This will remove the extra memory and time required to allocate the memory.

CC:  @jakirkham

mrocklin:
In general I agree that the main advantage here is probably for larger-than-memory data.  However, there are likely several things you can do to accelerating things on the Dask side here.  For example you might pass `name=False` to from_array to avoid that cost (it's hashing the data to find a deterministic key that you don't care about).  

@hmaarrfk might have some other suggestions.

I'm not sure if I'm reading the notebook correctly, but it seems like you have three approaches that perform as follows.  Is this right?

-  Numpy 40s
-  Dask array 3s
-  Some ITK specialized code 1.5s

If this is correct then I'm curious, what does the ITK specialized code do differently than Numpy?

hmaarrfk:
It seems to me you have a 6.7 GB dataset you are working on. I feel your pain. ITK is going at about 4GB/s. You are shrinking by `10, 10, 10` is that correct?

I've found numpy's code to be pretty slow and inefficient for large images. Slowdowns typically occur from a few places:

* Generous calls to `np.ascontiguousarray` which do wonders for small data (copying it is almost free) but which are very expensive for larger nd arrays.
* `np.pad` and `np.block` were optimized for small arrays where repeated concatenations is actually must faster than building up the metadata required for single shop padding/blocking. Please chime in here https://github.com/numpy/numpy/pull/11358 to express support as some believe that this is a "micro optimization". As a note, an `np.block` algorithm for large arrays just got in, so the performance of that one should be much better in 1.16 for arrays larger than  512x512.

[...]

PR #46, opened 2018-05-29
ENH: Add Dask Array basic support

thewtex:
Addresses task 1 from #44.

@mrocklin @dani-lbnl @jakirkham


itkwidgets/widget_viewer.py
```diff
@@ -39,6 +45,10 @@ def view(image):
     if isinstance(image, np.ndarray):
         image_from_array = itk.GetImageViewFromArray(image)
         viewer.image = image_from_array
+    elif have_dask and isinstance(image, dask.array.core.Array):
+        array = np.array(image)
+        image_from_array = itk.GetImageViewFromArray(array)
+        viewer.image = image_from_array
```
mrocklin: Two questions here:

1.  Is the passed in image the full dask array, or is this after something has sliced out the appropriate pieces from it
2.  Are there opportunities to do some of the preprocessing with dask ahead of time?
3.  I wonder if code like this could avoid the explicit type check, and instead rely on duck typing.  For example I wonder if it could be merged with the ndarray type check above with something like the following:

```python
def is_arraylike(x):
    return hasattr(x, 'shape') and hasattr(x, 'dtype') and hasattr(x, '__array__'):

if is_arraylike(x):
    x = np.asarray(x)
    image_from_array = itk.GetImageViewFromArray(image)
    viewer.image = image_from_array
```

This might both avoid the explicit dependence on dask.array and also open you up to other array-like objects beyond just dask.arrays.  (for more thoughts, see http://matthewrocklin.com/blog/work/2018/05/27/beyond-numpy)

thewtex: Great thoughts @mrocklin !

> Is the passed in image the full dask array, or is this after something has sliced out the appropriate pieces from it?

Correct, it is the full dask array, but Task 2, Task 3 of #44 would enable grabbing slices.

> Are there opportunities to do some of the preprocessing with dask ahead of time?

Yes, we could use dask for processing or daa or downsampling, possibly with `coarsen`, as recommended.

> I wonder if code like this could avoid the explicit type check, and instead rely on duck typing. 

Great idea. I'll take a look and test it out.

thewtex: Love the ideas in the blog. That should be a worthwhile efforts.

I tested your suggestion with `is_arraylike` and it works well. I made a minor tweak -- added `ndim`, which is needed in this case.

Thanks again for the review!

jakirkham:
Thanks for adding this, @thewtex. This should be fun to play with. 😄

thewtex:
@jakirkham should be fun and useful! A new version has been tagged and released, 0.10.0. Enjoy!

jakirkham:
Cool, thanks for doing that. Are there `conda` packages somewhere?

thewtex:
> Cool, thanks for doing that. Are there conda packages somewhere?

We could use some :-)

PR #92, opened 2018-10-10
ENH: Use ipydatawidgets for roi traitlet

thewtex:
@vidarf I am trying to use `ipydatawidgets` for a 2D float array that specifies a region of interest.

However, the change in the JavaScript model does not appear to be propagated to the Python kernel side. This is evident by the `viewer.roi_region()` still being all zeros in [this example notebook](https://github.com/InsightSoftwareConsortium/itk-jupyter-widgets/blob/master/examples/SelectRegionOfInterest.ipynb).

An ideas on what needs to change?

### DIFF
--- a/itkwidgets/widget_viewer.py
+++ b/itkwidgets/widget_viewer.py
@@ -13,6 +13,7 @@ import itk
 import numpy as np
 import ipywidgets as widgets
 from traitlets import CBool, CFloat, Unicode, CaselessStrEnum, Tuple, List, TraitError, validate
+from ipydatawidgets import NDArray, array_serialization, shape_constraints
 from .trait_types import ITKImage, itkimage_serialization
 try:
     import ipywebrtc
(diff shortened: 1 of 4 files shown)


vidartf:
For this usage scenario, I think it is easier to start with using the `NDArray` trait than `DataUnion`.

Then I would use the `simplearray_serialization`, which serializes to and from a JS object with the following signature:

```js
{
  shape: Array,
  array: TypedArray
}
```

E.g. the default value would be:
```js
{
  array: new Float64Array([0., 0., 0., 0., 0., 0.]),
  shape: [2, 3]
}
```

vidartf:
PS: For cases like this, it might make sense to have a "fixed shape" serializer on the JS side. I might implement that!

thewtex:
@vidartf That's perfect! :1st_place_medal: It works! Thank you!

`ipydatawidgets` looks very nice -- thanks for working on it!

PR #154, opened 2019-07-10
DOC: Add pyvista.PolyData example

thewtex:
@banesullivan I'm interested in your thoughts on how we can make this mesh rendering synergistic with pyvista.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -208,6 +208,7 @@ After installation, try the following examples that demonstrate how to visualize
 - `NumPy array point sets <https://github.com/InsightSoftwareConsortium/itk-jupyter-widgets/blob/master/examples/NumPyArrayPointSet.ipynb>`_
 - `ITK Mesh <https://github.com/InsightSoftwareConsortium/itk-jupyter-widgets/blob/master/examples/Mesh.ipynb>`_
 - `VTK vtkPolyData <https://github.com/InsightSoftwareConsortium/itk-jupyter-widgets/blob/master/examples/vtkPolyData.ipynb>`_
+- `pyvista PolyData <https://github.com/InsightSoftwareConsortium/itk-jupyter-widgets/blob/master/examples/pyvista.PolyData.ipynb>`_
 
 or how to:
 
--- a/examples/pyvista.PolyData.ipynb
+++ b/examples/pyvista.PolyData.ipynb
@@ -1,0 +1,115 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Install dependencies for this example\n",
+    "# Note: This does not include itk-jupyter-widgets, itself\n",
+    "import sys\n",
+    "!{sys.executable} -m pip install pyvista"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pyvista\n",
+    "from pyvista import examples\n",
+    "\n",
+    "from itkwidgets import view"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [
+    {
+     "data": {
+      "text/html": [
+       "\n",
+       "<table>\n",
+       "<tr><th>PolyData</th><th>Information</th></tr>\n",
+       "<tr><td>N Cells</td><td>2452</td></tr>\n",
+       "<tr><td>N Points</td><td>1335</td></tr>\n",
+       "<tr><td>X Bounds</td><td>1.391e+02, 1.655e+03</td></tr>\n",
+       "<tr><td>Y Bounds</td><td>3.209e+01, 1.320e+03</td></tr>\n",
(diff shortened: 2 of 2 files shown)


banesullivan:
This is interesting... REALLY interesting.

banesullivan:
@thewtex - this is major - Considering `vtkImageData` (`pyvista.UniformGrid`) is already supported by `itk-jupyter-widgets` and now `vtkPolyData` (`pyvista.PolyData`) is supported, this means we can pretty much plot anything using `itk-jupyter-widgets`... we'd just have to apply a `vtkExtractGeometry` filter (`.extract_geometry()` on all PyVista mesh types) and send that to `itk-jupyter-widgets`.

This is perfect timing with some other PRs in PyVista (see https://github.com/pyvista/pyvista/pull/268 and https://github.com/pyvista/pyvista/pull/300). I'm going to hack at this and see what I can make happen for prototypes and share those here in a bit.

*As for implementation and path forward:* I'm wondering if we should approach this the same way as the other notebook plotting tools we've added in PyVista including VTKjs through `panel`, X3D and K3D, which *convert* an already rendered VTK window or if we should approach this differently. Perhaps we could move forward with https://github.com/pyvista/pyvista/issues/299 and make a more general `BasePlotter` class and have a specific plotter class in PyVista that purely leverages `itk-jupyter-widgets` without having to render anything on the PyVista/VTK side.

@akaszynski - thoughts? This is absolutely the best option in my opinion for notebook plotting.

banesullivan:
@thewtex - I'm curious is there a way to instantiate a `Viewer` and add several `geometries` to it? Something along the lines of:

```py
import pyvista as pv
from pyvista import examples

from itkwidgets import view, Viewer

mesh = examples.download_st_helens().warp_by_scalar().extract_geomerty()

plotter = Viewer()
plotter.geometries = [mesh,]
plotter.geometries.append(mesh.contour())
plotter
```

This doesn't seem to be working... but I know I can pass many geometries upon instantiation, but ideally I'd like to make a class that holds on to a `Viewer` and adds data to that scene much like a PyVista plotter:

```py
class PlotterITK():
    def __init__(self, **kwargs):
        self.plotter = Viewer()
        
    def add_mesh(self, mesh, **kwargs):
        """Adds mesh to the scene"""
        mesh = mesh.extract_geometry()
        self.plotter.geometries.append(mesh)
        # Handle all the RGB mapping and styling as well
        
    def show(self, **kwargs):
        """Show in cell output"""
        return self.plotter
    
plotter = PlotterITK()
plotter.add_mesh(mesh) # pass args for coloring too
plotter.show()
```

banesullivan:
Also, I thought we could plot scalar attributes on the geometries... is this not currently possible?

banesullivan:
Also, what about plotting PolyData (`geometries`) along side ImageData:

```py
import pyvista as pv
from pyvista import examples

from itkwidgets import view

vol = examples.download_knee_full()

view(image=vol, geometries=vol.contour())
```

This seems to only show the image data/volume and not the geometries

thewtex:
@banesullivan sorry for the late follow-up -- I have just returned from the SciPy conference, and I am catching up with GitHub notifications :books: 



> This is interesting... REALLY interesting.

I am glad you think so -- I would love to see good integration with PyVista!



> @thewtex - this is major - Considering `vtkImageData` (`pyvista.UniformGrid`) is already supported by `itk-jupyter-widgets` and now `vtkPolyData` (`pyvista.PolyData`) is supported, this means we can pretty much plot anything using `itk-jupyter-widgets`... we'd just have to apply a `vtkExtractGeometry` filter (`.extract_geometry()` on all PyVista mesh types) and send that to `itk-jupyter-widgets`.

Yes, you nailed in right on the head! :building_construction:  :rocket: 



> This is perfect timing with some other PRs in PyVista (see [pyvista/pyvista#268](https://github.com/pyvista/pyvista/pull/268) and [pyvista/pyvista#300](https://github.com/pyvista/pyvista/pull/300)). I'm going to hack at this and see what I can make happen for prototypes and share those here in a bit.

Awesome!! Yes, it would be very cool to have these widgets as a plotter backend for PyVista. I do not know enough about the PyVista architecture to suggest a path forward, but it is worth noting that, like K3D, these widgets are based on the [ipywidgets / jupyter-widgets](https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20Basics.html), but they obviously map very closely to the PyVista / VTK data and rendering model.

thewtex:
> This doesn't seem to be working... I know I can pass many geometries upon instantiation, but ideally, I'd like to make a class that holds on to a `Viewer` and adds data to that scene much like a PyVista plotter:

Instead of

```
self.plotter.geometries.append(mesh)
```

does

```
self.plotter.geometries = self.plotter.geometries.append(mesh)
```

work? If not, I will certainly fix it :-)

This would be more in line with how the [ipywidgets traitlets](https://traitlets.readthedocs.io/en/stable/) work.

thewtex:
> Also, I thought we could plot scalar attributes on the geometries... is this not currently possible?

It will be possible very soon :-).

thewtex:
> Also, what about plotting PolyData (`geometries`) along side ImageData:

> This seems to only show the image data/volume and not the geometries

Thanks for the note -- #164 will track the fix for this.

banesullivan:
@thewtex - can I specify the name of geometry so that it's not just Geometry 0/1/2/etc.?

PR #256, opened 2020-01-15
Skeleton visualization with skan.csr.Skeleton support

thewtex:
@jni loved re-discovering skan to help with skeleton visualization! :heart: 

@phcerdan I wonder if SGEXT could be available in a Python package so we could also use it to perform the skeletonization?

Outputs:

image

image


phcerdan:
Nice! SGEXT has python bindings using pybind11 (at least 80% of the library), but I haven't generated python packages, nor uploaded them to pypi. It's indeed a priority, will let you know. Thanks for the ping!

jni:
@thewtex wow, this is so cool! :tada: :heart: :heart_eyes:

PR #285, opened 2020-03-06
ENH: Add support for JupyterLab 2

thewtex:
@martinRenou following your lead -- looks good?

### DIFF
--- a/js/package-lock.json
+++ b/js/package-lock.json
@@ -6129,9 +6129,9 @@
       "dev": true
     },
     "terser": {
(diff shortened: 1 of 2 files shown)


martinRenou:
I think it does yes! Does it still work with JupyterLab 1?
We had issues with JupyterLab `1.2.6` but I think it's fixed in `1.2.7`

thewtex:
I did observe some odd behavior about not being able to load extensions recently. Hopefully resolved with 1.2.7. Thanks for the review!

PR #288, opened 2020-03-18
:heavy_plus_sign: install itkwidgets

tkoyama010:
Hi all. Thank you for good example of jupyter-notebook. I think installation of itkwidgets is need in this example.

### DIFF
--- a/examples/pyvista.PolyData.ipynb
+++ b/examples/pyvista.PolyData.ipynb
@@ -11,7 +11,8 @@
     "# Install dependencies for this example\n",
     "# Note: This does not include itkwidgets, itself\n",
     "import sys\n",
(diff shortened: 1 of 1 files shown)


thewtex:
@tkoyama010 thanks for the kind feedback and pull request!

We have a comment noting in all the notebooks that it is assumed that `itkwidgets` is already installed. The reason for this is that adding the proposed line in the notebook will not immediately result in a working environment -- the user has to reload the browser tab to load the required JavaScript. This could be confusing to new users.

Perhaps we should also point to the installation instructions in the comments --  what do you think?

tkoyama010:
@thewtex Thank you for your reply.
> Perhaps we should also point to the installation instructions in the comments -- what do you think?
Yes you are right. I'll add comment.

PR #313, opened 2020-05-21
Require colorcet>=2.0.0

### LINKED ISSUE #308: install_requires needs a specific version of colorcet
Getting an issue when using `view` with `pyvista`:

```
/opt/conda/lib/python3.7/site-packages/itkwidgets/widget_viewer.py in _validate_point_set_colors(self, proposal)
    456         if len(value) < n_colors:
    457             for index in range(len(value), n_colors):
--> 458                 color = colorcet.glasbey[index % len(colorcet.glasbey)]
    459                 result[index, :] = matplotlib.colors.to_rgb(color)
    460         return result

AttributeError: module 'colorcet' has no attribute 'glasbey'
```

Turns out you need a fairly recent version of `colorcet` and the current `install_requires` doesn't specify a version.

akaszynski:
Simple fix to ensure `colorcet` is the correct version.

Resolves #308

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -153,7 +153,7 @@ setup_args = {
         ]),
     ],
     'install_requires': [
-        'colorcet',
+        'colorcet>=2.0.0',
         'itk-core>=5.1.0',
         'itk-filtering>=5.1.0',
         'itk-meshtopolydata>=0.6.0',


thewtex:
@akaszynski thank you! :pray:

Issue #304, opened 2020-05-08
Piecewise function traits


thewtex:
Add traits for the Gaussians that specify the piecewise functions that define volume rendering opacity.

Issue #273, opened 2020-01-29
Question: output the results of the histogram manipulation


Sh4zKh4n:
This is more of a question. When manipulating the histogram for max and minimum values. Is there a way to output the new histogram as a file or result? It would help in quicker processing of data by reviewing the data set quickly to then manipulate the image for processing in a pipeline? 

If any one has any thoughts, id be interested to hear them?

thewtex:
@Sh4zKh4n are you referring to the opacity transfer function settings? The ability to store / restore these may come in the future.

Currently, it is possible to specify the min / max by passing `vmin` and `vmax` arguments to `view`.

Sh4zKh4n:
Hi Sorry, I was ill for a bit and had to take a break. The  ticket for #328 may be what I am suggesting. If you can read the file outside of itk i.e. plot the histogram from the out put data. For example, if I have a massive data set and I do a histogram of this, do some manual clipping of the histogram or stretching.  I do some further work and realise, i didnt save the fig but want to explore that histogram further or do some processing of the raw data to fit that itk based manual processing. Would there be a way to output the results to a file in array or dataset file. Then use those outputs with any computation. Sorry about the really slow reply, Ive only really come back to this.

ZhiyLiu:
Hi. [itk.LabelImageToStatisticsLabelMapFilter](https://itk.org/Doxygen/html/classitk_1_1LabelImageToStatisticsLabelMapFilter.html) will do the trick. Additionally, I would suggest to refer this [article](http://www.insight-journal.org/browse/publication/176) for the definition of itk.LabelObject. 

Simply put, the result from LabelImageToStatisticsLabelMapFilter contains histogram informations (e.g., frequency, bins). Based on these information you can draw the histogram with any library (e.g., matplotlib) you prefer.

PR #328, opened 2020-06-12
Opacity function trait

thewtex:


thewtex:
Closes #304 
Closes #273

PR #353, opened 2020-06-29
Treebeard

thewtex:


thewtex:
@alex-treebeard @Rabscuttler please take a look

.github/workflows/notebook-test.yml
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    name: Run notebooks on treebeard
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - uses: treebeardtech/treebeard@master
+        with:
+          docker-username: thewtex
```
alex-treebeard, line 14: Seems like for your use case you can delete treebeard.yaml and just use the action inputs to specify which notebooks you want to run using `notebooks: "examples/3DImage.ipynb"` inside the `with:` section.

`notebooks:` in treebeard.yaml is useful for multiple glob patterns (you can't pass lists into github actions!)

You can see all action inputs here: https://github.com/treebeardtech/treebeard/blob/master/action.yml

RE the other fields in treebeard.yaml:

`output_dirs` is only necessary for uploading outputs to treebeard teams -- I'd say we should not bother with that just yet.

`ignore` should just be used for ignoring notebooks you don't want to run.

`schedule` is deprecated -- please remove it as it's dangerous copypasta :)

thewtex: @alex-treebeard thanks for the tips! :pray: I'll remove *treebeard.yaml*, and use `with:` and `notebooks:`.

PR #427, opened 2021-07-16
DOC: Use Jupyter Notebook binder paths

thewtex:
Move back to JupyterLab after we have JupyterLab 3 support


aylward:
These work great!   Very nice!!!

Issue #16, opened 2018-03-20
Add documentation


thewtex:
Sphinx / ReadTheDocs

https://ipywidgets.readthedocs.io/en/latest/embedding.html#embedding-widgets-in-the-sphinx-html-documentation

thewtex:
PyData Theme:

https://pydata-sphinx-theme.readthedocs.io/en/stable/index.html

and jupyterlite-sphinx installed:

https://jupyterlite-sphinx.readthedocs.io/en/latest/

thewtex:
Draft documentation sections:

- Quick start Guide
- Deployments (name??)
  * Colab
  * Jupyter
  * Jupyterlab
  * VSCode
  * Jupyterlite
  * Server app
  * Desktop app

- Integrations
  * numpy
  * itk
  * vtk
  * pytorch
  * monai
  * dask
  * xarray
  * pyvista
  * vedo
  * pyimagej
  * skan
  * zarr

thewtex:
:tada: https://itkwidgets.readthedocs.io/en/latest/

PR #456, opened 2022-06-30
Docs

bnmajor:
Addresses #16


bnmajor:
@thewtex This covers most of the topics covered in #16 but there are some pieces that are missing. I need some clarification on what the following sections should be actually be covering:

- Deployments
  - Server app
  - Desktop app
- Integrations: Should this just be covering installation and maybe an example of using that dependency?
- Jupyterlite and VSCode are not yet covered in the docs as there are still issues with both that I am trying to sort out. VSCode is not displaying the viewer (although it also does not produce errors and works perfectly in Jupyter) and Jupyterlite has issues with installation as it requires pure python wheels for all packages (must be installed with `piplite`):
  ```
  ValueError: Couldn't find a pure Python 3 wheel for: 'multidict<5.0,>=4.5', 'yarl<2.0,>=1.0'
  ```

thewtex (approved):

thewtex:
>  I need some clarification on what the following sections should be actually be covering

Let's cover the sections we do not have content for in a follow-up PR.

> Integrations: Should this just be covering installation and maybe an example of using that dependency?

:+1: both :-)

docs/deployments.rst
```diff
@@ -0,0 +1,67 @@
...
+To use itkwidgets locally first `install Jupyter Notebook`_ and start Jupyter:
+
+.. _install Jupyter Notebook: https://jupyter.org/install#jupyter-notebook
+
+.. code-block::
+
+    pip install notebook
+    jupyter notebook
+
+If you'd rather interact with reomtely hosted notebooks you can also open them
```
labdala, line 39: small detail "remotely"

labdala (approved):

thewtex (approved):
Loving the screeshots! :heart_eyes:

bnmajor:
@thewtex This adds the integrations section as well as screenshots, support for the listed integrations, and new notebook examples. Two things to address though:
1. I do not have permissions to add a webhook to get `readthedocs` set up
2. Was the zarr example notebook that you added working for you? I was able to download the data locally and test it to get the screenshot but actually running the notebook does not work for me (I can open a separate issue/PR for this if this is a bug)

docs/deployments.rst
bnmajor: Thanks for catching this!

thewtex:
> I do not have permissions to add a webhook to get readthedocs set up

Could you please add my account to the ReadTheDocs project?

> Was the zarr example notebook that you added working for you? I was able to download the data locally and test it to get the screenshot but actually running the notebook does not work for me (I can open a separate issue/PR for this if this is a bug)

Yes, it was -- please create an issue.

Thanks!

bnmajor:
> > I do not have permissions to add a webhook to get readthedocs set up
> 
> Could you please add my account to the ReadTheDocs project?

I added you as a maintainer so you should be able to access it now: https://readthedocs.org/projects/itkwidgets/. If not let me know, there may be additional steps I need to take.

> > Was the zarr example notebook that you added working for you? I was able to download the data locally and test it to get the screenshot but actually running the notebook does not work for me (I can open a separate issue/PR for this if this is a bug)
> 
> Yes, it was -- please create an issue.

Great will do, thanks!

bnmajor:
> Thanks. I enabled the GitHub Integration. It should be working now.
Thanks! I think everything is all set with that now, docs should build on pull requests and the default branch is set to main.

PR #458, opened 2022-07-04
ENH: Embed screenshot in output

### LINKED ISSUE #445: HTML output in the notebook
Generate a base64-encode image from the screenshot feature of the viewer after the widget is output and the scene is rendered and place it in the notebook HTML output. Make it invisible with JavaScript as the interactive viewer is used, but it should exist on the saved notebook output so it can be seen in static contexts, like browsing notebooks on GitHub.

bnmajor:
Display the embedded image in static views of the notebook, otherwise display the viewer.

Fixes #445


thewtex (changes requested):
Todos:

- [x] Run on additional notebooks
- [x] Initial screenshot capture, then capture after delay
- [ ] Image updated when the Screenshot button is pressed

bnmajor:
This branch relies on Kitware/itk-vtk-viewer#520 for the "update screenshot on icon click" feature

thewtex (approved):
:100:

thewtex:
:star2:

PR #468, opened 2022-07-11
Version deps

thewtex:
@tbirdso thanks for the earlier patch, we have to backtrack here a bit, but we can revert after 5.3.0 is out.


tbirdso:
No worries @thewtex , is there an issue that could be linked to help me understand the reason?

thewtex:
@tbirdso want to get more testing on the itk pre-release packages.

PR #471, opened 2022-07-18
DOC: Add instructions for jupter lab extension installation

bnmajor:
In trying a fresh environment it seems that the `jupyter` command is required in order to install the ImJoy extension for JupyterLab.


bnmajor:
This instruction is not necessary, closing.

PR #472, opened 2022-07-18
feat(ChooseUI): Add initialization option to specify the UI

labdala:
**Under Construction - not ready to merge**

The variable 'whichUI' should be used to specify which UI itk-vtk-viewer should use. In case 'whichUI=pydata-sphinx', then itk-vtk-viewer should use the UI JS module from itk-viewer-bootstrap-ui on jsdelivr.


labdala:
The idea was to add this in the main branch. See #482 .

PR #474, opened 2022-07-20
Image label

bnmajor:
Add support for labels

Relies on Kitware/itk-vtk-viewer#522

### DIFF
--- a/itkwidgets/integrations/__init__.py
+++ b/itkwidgets/integrations/__init__.py
@@ -11,7 +11,7 @@ from ..render_types import RenderType
 
 _image_count = 1
 
-async def _set_viewer_image(itk_viewer, image, name=None):
+async def _set_viewer_image(itk_viewer, image, name=None, is_label=False):
     global _image_count
     if isinstance(image, itkwasm.Image):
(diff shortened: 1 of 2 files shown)


bnmajor:
This is currently still marked as a draft because in some of my testing I am seeing the label rendering take so long it is essentially unusable (~4-5 minutes). In other cases I am not seeing this issue though (it may be seconds slow but not minutes) but in both cases it seems to spend most of the time in [`toMultiscaleSpatialImage`](https://github.com/Kitware/itk-vtk-viewer/blob/master/src/IO/toMultiscaleSpatialImage.js#L39-L82).

PaulHax:
> label rendering take so long it is essentially unusable (~4-5 minutes)

itk-vtk-viewer state machine can get loopy off the golden path.  I'm happy to try to reproduce in JS on https://github.com/Kitware/itk-vtk-viewer/pull/522 if you like.

thewtex:
@PaulHax yes, please take a look :pray:

bnmajor:
@PaulHax Thanks, I appreciate it!

PR #476, opened 2022-07-20
Notebook updates

bnmajor:
Some notebooks made assumptions about environment or had old debug information. This PR removes debug cells and ensures we always install the correct version of `itkwidgets`. This also removes references to `itk.GroupSpatialObject`. This is not raising errors in testing in Jupyter Notebook or JupyterLab, however for some reason this errors out in Colab an prevents itkwidgets from being used.


thewtex (approved):
@bnmajor thank you!

PR #479, opened 2022-07-21
DOC: Simplify installation instructions with package extras

thewtex:

PR #480, opened 2022-07-21
ENH: Include imjoy-elfinder in options

bnmajor:
Addresses part of #479 by adding `imjoy-elfinder` dependency to options


thewtex (approved):
:100:

PR #483, opened 2022-07-29
DOC: Remove mention of requirements-dev.txt

aylward:
The requirements-dev.txt file has been removed, so mention of it should be removed from documentation.


thewtex (approved):
@aylward thanks!

PR #488, opened 2022-08-01
Label fixes

bnmajor:
- Fixes another bug that was introduced by the incorrectly resolved merge conflicts. Support for the `label_image` input was lost, added back in now.
- Adds the `label_image` info to the doc string
- Updates the MONAI example notebook to use `label_image`

Relies on the changes in Kitware/itk-vtk-viewer#538


labdala (approved):
Looks great!

PR #491, opened 2022-08-02
fix(init-input-data): Fix arguments for _get_viewer_point_sets

labdala:


bnmajor (approved):
So many things resolved incorrectly! Thank you for the find and fix!

PR #494, opened 2022-08-08
BUG: Fix uninitialized variable in viewer

### LINKED ISSUE #398: View does not display point sets by default
### Overview
Trying to display an ITK 2D point set with "view" fails to show data with the "points" representation. The itkwidgets UI overlay can be used to set points to "sphere" representation which allows data to be displayed. An uncaught TypeError is present in the console.

    at initialize_itkVtkViewer (viewer.js:932)
    at initialize_viewer (viewer.js:1852)
    at O (viewer.js:286)
    at viewer.js:1285

### Steps to Reproduce
1. Generate an ITK point set of type `itk.PointSet[itk.F,2]`
2. Create a itkwidgets viewer object in Jupyter Notebook with `view(point_sets=[point_set])`

### Expected Behavior: 
The point set "points" representation is visible.

### Observed Behavior: 
The "points" representation is not visible. The "spheres" representation can be manually selected to make points visible.

### Other notes: 
Bug not observed in example using numpy arrays instead of ITK point sets: [https://github.com/InsightSoftwareConsortium/itkwidgets/blob/master/examples/NumPyArrayPointSet.ipynb](https://github.com/InsightSoftwareConsortium/itkwidgets/blob/master/examples/NumPyArrayPointSet.ipynb)

Erroneous behavior can be seen in [ITKExamples Draft PR 189](https://github.com/InsightSoftwareConsortium/ITKExamples/pull/189)

### LINKED ISSUE #429: geometery_colors, other properties not working when passed
On my installlations (and on binder right now), passing some initial settings for rendering do not work. 

For example, passing geometry_colors=['g'] does not cause the Mesh example to initially render green.

```
viewer=view(geometries=mesh, geometry_colors=['g'] , ui_collapsed=True)
```

However, if you start the viewer and then display it and then set the colors as an np.array

```
viewer.geometry_colors = np.array([[0,1,0]], dtype=np.float32)
```

then the colors do change dynamically.  But you can't set them to start (everything is always red).. 

checking the viewer.get_state('geometry_colors') indicates that the front end did receive the correct data when passing geometry_colors to the view function, but it is not properly syncing what is rendered.

okmiim:
## Error

Browser developer console:
```
Uncaught (in promise) TypeError: channels is null
    initialize_itkVtkViewer viewer.js:933
    initialize_viewer viewer.js:1852
    createRenderingPipeline viewer.js:286
    render viewer.js:1285
viewer.js:933:8
```

Problems:
- See #398, #429 
- `point_set_colors` does not work
- ...

## Steps to reproduce

1. Open e.g. `examples/NumPyArrayPointSet.ipynb`
2. Open developer console of the browser
3. Run the notebook
4. Look in the console
   1. Error message in console
   2. All point clouds are white (normally they have different colors)

## Fix

This just adds a null check to the `channels` attribute.
Additionally, a similar case is also mitigated (`background`).

Fixes #398
Fixes #429

### DIFF
--- a/js/lib/viewer.js
+++ b/js/lib/viewer.js
@@ -232,7 +232,7 @@ const createRenderingPipeline = (
     backgroundColor[2] = rgb[2] / 255.0
   }
   const backgroundTrait = domWidgetView.model.get('background')
-  if (backgroundTrait.length !== 0) {
+  if (backgroundTrait && !!backgroundTrait.length) {
     backgroundColor = backgroundTrait
   }
   const viewerStyle = {
@@ -721,7 +721,7 @@ const ViewerView = widgets.DOMWidgetView.extend({
     }
     this.model.itkVtkViewer.on('backgroundColorChanged', onBackgroundChanged)
     const background = this.model.get('background')
-    if (background.length === 0) {
+    if (background === null|| background.length === 0) {
       this.model.set('background', this.model.itkVtkViewer.getBackgroundColor())
     }
 
@@ -930,7 +930,7 @@ const ViewerView = widgets.DOMWidgetView.extend({
       onChannelsChanged
     )
     const channels = this.model.get('channels')
-    if (channels.length === 0) {
+    if (channels === null || channels.length === 0) {
       this.model.set('channels', this.model.itkVtkViewer.getComponentVisibilities())
     }
 


thewtex (approved):
@okmiim thank you for the contribution! :pray:

thewtex:
Failing CI unrelated

PR #500, opened 2022-08-09
REL: Do not include examples or node_modules in release

### LINKED ISSUE #496: Ignore node_modules/ when creating sdist
The sdist created with `hatch build` is 101 MB -- we do not want to bundle a local `node_modules/`.

bnmajor:
Fixes #496


thewtex (approved):
@bnmajor thank you!

PR #507, opened 2022-08-14
Update readme

bnmajor:
This adds installation instructions to the README since JuypterLab and Jupyter Notebook require the ImJoy extension and anyone installing from main but used to the previous install steps may not realize this (and the error message or just blank output cell does not make it obvious what is going wrong).


thewtex (approved):
@bnmajor thanks! :book:

PR #527, opened 2022-09-20
ENH: Server-side multi-scale chunking of images

thewtex:


itkwidgets/integrations/__init__.py
```diff
@@ -1,60 +1,132 @@
...
     if HAVE_VTK:
         import vtk
         if isinstance(point_sets, vtk.vtkPolyData):
             return vtk_polydata_to_vtkjs(point_sets)
-    if HAVE_DASK:
-        import dask
-        if isinstance(point_sets, dask.array.core.Array):
-            return dask_array_to_ndarray(point_sets)
+    if isinstance(point_sets, dask.array.core.Array):
+        return dask_array_to_ndarray(point_sets)
```
bnmajor, line 121: It looks like `dask_array_to_ndarray` has been removed. Should we keep it around for this case or just move the functionality here?

bnmajor:
@thewtex This is so awesome!! When I run this locally with Jupyter Lab it seems to work great :slightly_smiling_face: 

Running with SageMaker gives me the following error:
```
Traceback (most recent call last):
  File "/home/studio-lab-user/.conda/envs/default/lib/python3.9/site-packages/imjoy_rpc/rpc.py", line 404, in _wait
    result = await result
  File "/home/studio-lab-user/.conda/envs/default/lib/python3.9/site-packages/itkwidgets/viewer.py", line 82, in run
    result = _get_viewer_image(data)
  File "/home/studio-lab-user/.conda/envs/default/lib/python3.9/site-packages/itkwidgets/integrations/__init__.py", line 67, in _get_viewer_image
    multiscale = itk_image_to_multiscale(image, scale_factors=scale_factors)
  File "/home/studio-lab-user/.conda/envs/default/lib/python3.9/site-packages/multiscale_spatial_image/to_multiscale.py", line 715, in itk_image_to_multiscale
    from itk_helper import xarray_from_image
ModuleNotFoundError: No module named 'itk_helper'
```

When I try it in Colab I get:
```
ERROR: Could not find a version that satisfies the requirement multiscale_spatial_image==0.10.0 (from versions: 0.5.0, 0.6.0, 0.7.0, 0.8.0, 0.9.0)
ERROR: No matching distribution found for multiscale_spatial_image==0.10.0
```
It seems that Colab is running Python 3.7 and `multiscale_spatial_image` requires >=3.8.

thewtex:
> ModuleNotFoundError: No module named 'itk_helper'

Bump the dependency version to 0.10.1 to address this.

> It seems that Colab is running Python 3.7 and multiscale_spatial_image requires >=3.8.

Yes, we have a dependency we need to backport to Python 3.7 -- on the list! 🔢

itkwidgets/integrations/__init__.py
```diff
@@ -1,60 +1,132 @@
...
+    if multiscale is None:
+        if isinstance(image, zarr.Array):
+            spatial_image = to_spatial_image(image)
+            scale_factors = _spatial_image_scale_factors(spatial_image, min_length)
+            multiscale = to_multiscale(spatial_image, scale_factors)
+    if multiscale is None:
+        # NGFF Zarr
+        if isinstance(image, zarr.Group) and 'multiscales' in image.attrs:
+            return image.store
+    if multiscale is None and HAVE_TORCH:
```
PaulHax, line 88: It is worth removing the `is multiscale` checks with a `func makeMultiscale(image)` ?
```
def _get_viewer_image(image):

PaulHax (approved):
Examples work locally for me.  Chunk all the things.

itkwidgets/integrations/__init__.py
thewtex: Good catch. Let's just use `np.asarray` here for now.

itkwidgets/integrations/__init__.py
thewtex: Good idea -- yes, that is a bit cleaner.

PR #530, opened 2022-09-20
chore: add byte compiled and venv to git ignore

PaulHax:
Makes repo a little more approachable for folks without python stuff in their global .gitignore?


thewtex (approved):
@PaulHax thanks!

PR #557, opened 2022-10-20
Colab set image

bnmajor:
When running a notebook in Colab the viewer was failing on `to_ngff_zarr`. It seems that in Colab specifically there are issues using Dask from an asynchronous function. By moving this to the (now) synchronous setup function we avoid this issue.

This branch also adds Colab badges to the notebooks.

### DIFF
--- a/examples/Hello3DWorld.ipynb
+++ b/examples/Hello3DWorld.ipynb
@@ -1,6 +1,16 @@
 {
  "cells": [
   {
+   "cell_type": "markdown",
+   "id": "23d710b3-dd78-4f76-895f-7e22f194d738",
+   "metadata": {},
+   "source": [
+    "### Try this notebook in Google Colab as well!\n",
+    "\n",
+    "[![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/InsightSoftwareConsortium/itkwidgets/blob/main/examples/Hello3DWorld.ipynb)"
+   ]
+  },
+  {
    "cell_type": "code",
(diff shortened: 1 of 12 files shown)


thewtex:
@bnmajor I am wondering if we can push the dataset conversion / `to_ngff_zarr` up even further into the `Viewer` class?

Perhaps we can keep `setup` `async`. Maybe this could help with #564

bnmajor:
@thewtex Digging into this a bit more I find that with a minimal plugin in a notebook:
```
class ImJoyPlugin1():
    async def setup(self):
        pass

    async def run(self, ctx):
        # create a viewer
        viewer = await api.createWindow(src="https://bafybeicuvbx4cfp3zgo25j376bmmysh73a2ui5dbapuwqtmjcvcaipmpbi.on.fleek.co/")
        # show an image
        viewer.setImage(image)

class ImJoyPlugin2():
    async def setup(self):
        pass

    async def run(self, ctx):
        # create a viewer
        viewer = await api.createWindow(src="https://bafybeicuvbx4cfp3zgo25j376bmmysh73a2ui5dbapuwqtmjcvcaipmpbi.on.fleek.co/")
        # show an image
        viewer.setImage(image)

api.export(ImJoyPlugin1())
api.export(ImJoyPlugin2())
```
Only the last viewer will be created and shown. `ImJoyPlugin1` `setup()` and `run()` are never even reached so it seems that multiple calls to the imjoy-rpc api interrupt any previous calls, so unfortunately pushing the data conversion up to the `Viewer` class does not help this bug.

PR #558, opened 2022-10-20
Jupyterlite

bnmajor:
Fixes the code path for now to avoid background jobs/threading in JupyterLite.


itkwidgets/integrations/environment.py
```diff
@@ -0,0 +1,24 @@
...
+    try:
+        from IPython import get_ipython
+        parent_header = get_ipython().parent_header
+        username = parent_header['header']['username']
+        if username == '':
+            IN_JUPYTERLAB = True
+        elif username == 'username':
+            IN_JUPYTER_NOTEBOOK = True
+        else:
+            IN_AWS = True
```
bnmajor, line 22: This condition has not been tested as I am still unable to access Studio Lab, all other conditions have.

itkwidgets/integrations/environment.py
```diff
@@ -0,0 +1,24 @@
+IN_JUPYTER_NOTEBOOK = False
+IN_JUPYTERLAB = False
+IN_JUPYTERLITE = False
+IN_AWS = False
```
PaulHax, line 4: If there can only be 1 as true, an Enum would enforce that.
```
def findEnv():
  if something:
    return Env.JUPYTERLAB
  if anotherThing:
   return Env.JUPYTERLITE

env = findEnv()
```

thewtex (changes requested):
Awesome!!

itkwidgets/integrations/environment.py
bnmajor: Updated, thanks!

itkwidgets/integrations/environment.py
```diff
@@ -0,0 +1,24 @@
...
+        parent_header = get_ipython().parent_header
+        username = parent_header['header']['username']
+        if username == '':
+            IN_JUPYTERLAB = True
+        elif username == 'username':
+            IN_JUPYTER_NOTEBOOK = True
+        else:
+            IN_AWS = True
+    except AttributeError:
+        IN_JUPYTERLITE = True
```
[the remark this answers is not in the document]
bnmajor: Fixed!

itkwidgets/viewer.py
```diff
@@ -101,11 +103,13 @@ async def run(self, ctx):
...
-        itk_viewer.registerEventListener(
-            'renderedImageAssigned', self.set_event
-        )
-        # Once the viewer has been created any queued requests can be run
-        asyncio.get_running_loop().call_soon_threadsafe(self.viewer_event.set)
+        if not IN_JUPYTERLITE:
+            itk_viewer.registerEventListener(
+                'renderedImageAssigned', self.set_event
+            )
+        if not IN_JUPYTERLITE:
```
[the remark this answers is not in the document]
bnmajor: Oops! Yes, thank you, fixed now...

thewtex (approved):
Awesome, awesome!

PR #559, opened 2022-10-20
ENH: Add 'all' package extras and use in examples

thewtex:
A default that 'just works' across all environments.


bnmajor (approved):
Love this!

PR #569, opened 2022-10-25
feat(viewer): no interpolation on labels, categorical color maps

### LINKED ISSUE #565: If a label_image is loaded, Interpolation should be set to False.
If interpolation is applied to a label_image, then values not in the label_map appear in the visualization - making labels appear to be incorrect or undefined at their edges.

PaulHax:
Fixes for transfer function editor

closes #565


thewtex (approved):
Thanks @PaulHax !

PR #570, opened 2022-10-27
Warnings and errors

bnmajor:


itkwidgets/integrations/environment.py
```diff
@@ -34,3 +35,15 @@ def find_env():
 
 
 ENVIRONMENT = find_env()
+
+
+output = subprocess.getoutput('pip list | grep imjoy')
```
thewtex, line 40: For portability and efficiency, we want to use `try: import <package>`

thewtex (changes requested):
@bnmajor thanks!

Tweak to detecting dependencies noted inline.

itkwidgets/integrations/environment.py
bnmajor: Sure thing! I was also going to update the message for Jupyterlite since it's not a `pip install` fix, but it's also not a simple `piplite.install` either (trying `await piplite.install('imjoy-jupyterlab-extension')` returns `ValueError: Can't find a pure Python 3 wheel for 'tornado>=6.1.0'`. Should the error just be something like "try piplite.install... or raise an issue: [link to new issue form]" (or something along those lines) or just "piplite.intsall" or something else?

thewtex: Yes, lot's just recommend that they install the package of the given name -- conda/mamba could be used to install the package, too.

thewtex (approved):
:100: :1st_place_medal:

PR #583, opened 2022-11-17
BUG: Install all package dependencies in CI

thewtex:
Avoid recent environment dependency check failures.


thewtex:
@bnmajor looking to fix current CI failures on `main`

bnmajor:
@thewtex  Makes sense, thanks! Looks like my logic is flawed though when raising errors though. I check if we're in Colab but we also need to make sure we're in Python >= 3.8 before warning as well. I can put in a fix for that

bnmajor (approved):
Awesome, thank you!

Issue #568, opened 2022-10-25
Need to make getters available


bnmajor:
We should expose the getter functions to users so that they can access the information that they may need. Currently progress on this is barred by two issues, the first of which being the major barrier:

- (Highest Priority) For some reason we currently get stuck with awaiting the results. Calling the [ImJoyPluginAPI](https://github.com/Kitware/itk-vtk-viewer/blob/master/src/ImJoyPluginAPI.js) we can see that the value is retrieved, but for some reason the `<FuturePromise pending>` never resolves when we `await` it. Can be reproduced with the following:
```
from imjoy_rpc import api
import imageio.v2 as imageio

image = imageio.imread("https://data.kitware.com/api/v1/file/5b843d468d777f43cc8d4f6b/download/engine.nrrd")

class ImJoyPlugin():
    async def setup(self):
        pass

    async def run(self, ctx):
        # create a viewer
        viewer = await api.createWindow(src="https://bafybeicuvbx4cfp3zgo25j376bmmysh73a2ui5dbapuwqtmjcvcaipmpbi.on.fleek.co/")
        # show an image
        viewer.setImage(image)

    async def getUICollapsed(self):
        return await self.viewer.getUICollapsed()

viewer = ImJoyPlugin()
api.export(viewer)
await viewer.getUICollapsed()
```
- (Useful but lower priority) To handle the case of "Run All" we need to think about how to handle calls to these getters so that they don't produce errors that interrupt the notebook. Some ideas:
  - Use the background thread like we do with the setters. In this case we would need to rethink the current approach a little to make sure we get the result back. We'd also need a way to associate the result with the correct cell. Would also need a "placeholder" for the returned value to indicate that it hasn't yet returned even though the cell has "completed". Less than ideal.
  - Try to move the rpc to the background thread so that cells can block while they wait for results.

bnmajor:
Linking [this discussion](https://github.com/imjoy-team/imjoy-rpc/issues/449) here to keep this issue updated.

thewtex:
Related:

- https://github.com/Kirill888/jupyter-ui-poll
- https://github.com/kafonek/ipython_blocking

bnmajor:
Fixed in #587 and improved in #702. Closing.

PR #587, opened 2022-12-14
Add support for getters

bnmajor:
-- WIP --

Guided by the items discussed in #568 this branch attempts to add support for getter functions. Adapting approaches in the [ipython_blocking](https://github.com/kafonek/ipython_blocking) and [jupyter-ui-poll](https://github.com/Kirill888/jupyter-ui-poll) libraries the general approach is to:

- Temporarily overwrite execute_request handler in IPython in order to collect and queue requests
- Once the viewer is created the background thread calls `execute_next_request` to start the cycle of processing the remaining cells
- Pop requests off one at a time and inspect for getters
  - If getters are found we create a future and attach a callback that sets the result once it is received. Once all results have resolved the cell is re-run so that it may return the results.
  - If no getters are found the request is just run as usual

Issues:
- Currently the user namespace is not being updated when the cell is re-run properly so cell output is always wrong


itkwidgets/viewer.py
```diff
@@ -227,42 +255,107 @@ def set_image(self, image: Image):
...
+        elif self.results['getImageGradientOpacityScale'].done():
+            return self.results.pop('getImageGradientOpacityScale').result()
 
     def set_image_interpolation_enabled(self, enabled: bool):
         self.queue_request('setImageInterpolationEnabled', enabled)
+    def get_image_interpolation_enabled(self):
+        if 'getImageInterpolationEnabled' not in self.results:
+            self.call_getter('getImageInterpolationEnabled')
+        elif self.results['getImageInterpolationEnabled'].done():
+            return self.results.pop('getImageInterpolationEnabled').result()
```
thewtex, line 326: Can we put this sequence of calls in a decorator?

bnmajor: Yup, updated!

bnmajor:
This approach is making progress but is still a bit naive and has a lot of additional cases/needs to take into account. Starting a list here of key concerns and shortcomings so that they are addressed as this moves forward:

- [ ] We are currently running getters twice - once before and once during cell execution. This approach needs to be re-thought as there are plenty of use cases where this would not return the correct value. ex:
- [ ]

bnmajor:
This approach is making progress but is still a bit naive and has a lot of additional cases/needs to take into account. Starting a list here of key concerns and shortcomings so that they are addressed as this moves forward:

- [ ] We are currently running getters twice - once before and once during cell execution. This approach needs to be re-thought as there are plenty of use cases where this would not return the correct value. ex:
```
print(viewer.get_axes_enabled()) # False by default
viewer.set_axes_enabled(True)
print(viewer.get_axes_enabled()) # Should be True, will return False
```
- [ ] Currently using regex to identify getters. There are plenty of cases where a getter or setter may not be called directly so this is not a viable approach.
- [ ] Need to confirm that none of the approaches that we settle on inadvertently affect the notebook history or output

thewtex:
Todos

- Multiple viewers
- Test notebook: 1) set, get, throws exceptions
- Intercept getters during execution instead of after

which should help with all three points above.

-> merge

Create another issue as needed for setting cell output with a getter output.

thewtex:
@bnmajor this timeout only occurs when the tests are executed through nbmake?

@alex-treebeard are notebook events processed in a different way when executed through nbmake?

thewtex:
@alex-treebeard suggested that there may be a to resolve via the way nbmake uses [nbclient](https://nbclient.readthedocs.io/en/latest/reference/nbclient.html).

bnmajor:
@thewtex So it seems like the `nbmake` issue is not actually with the getters/setters but with the fact that in testing the itk-viewer is never actually created (so any cells with getters/setters timeout waiting for it to be created before being run). If I use the  `--overwrite` flag to see the results produced from testing there is no embedded screenshot which seems to back-up this theory....

thewtex:
@bnmajor oh, good catch! :bulb: 

[Indeed](https://nbclient.readthedocs.io/en/latest/client.html?highlight=browser#widget-state)

> This widget rendering is not performed against a browser during execution

@oeway any ideas / suggestions / pointers to have ImJoy talk to a headless chromium browser?

alex-treebeard:
@thewtex been doing a bit of reading around widgets. Is it necessary to load the browser in order to initialise the ITK widget?

What needs to run in order to unblock the getters from executing against an initialised viewer object?

thewtex:
@alex-treebeard 

> Is it necessary to load the browser in order to initialise the ITK widget?

It is not necessary to import the Python code, run it. However, a browser is necessary for the graphical context and JavaScript environment to perform renderings and work with the client-side widget state.

> What needs to run in order to unblock the getters from executing against an initialised viewer object?

One way or another, we need to be talking to, e.g., Chrome / Electron in Headless Mode, when running in nbmake / nbclient.

alex-treebeard:
@thewtex have you checked out [galata](https://github.com/jupyterlab/jupyterlab/tree/master/galata)?

It's possible to stand up a jupyter lab instance in a test environment, then we could potentially run nbmake against its kernel.

thewtex:
@alex-treebeard great idea!!

Yes, we could use a version of `nb_run.py` that uses galata and something along the lines of:

https://github.com/jupyterlab/jupyterlab/blob/2250b956294558de58a99e4a26286270654eb416/galata/test/galata/notebook.spec.ts#L91-L115

thewtex:
Let's merge this, then, in separate PR's

- Document getters limitations in sphinx
- Address the JS testing issues

thewtex (approved):

thewtex (approved):

PR #589, opened 2022-12-22
Cli

thewtex:


thewtex:
@bnmajor this is scratch WIP, related:

- [hypha](https://github.com/imjoy-team/hypha)
- [imjoy-rpc hypha](https://github.com/imjoy-team/imjoy-rpc/tree/master/javascript/src/hypha) support

CC @oeway

thewtex:
@bnmajor could latest updates please be pushed?

bnmajor:
> @bnmajor could latest updates please be pushed?

@thewtex Updated!

thewtex:
@bnmajor @oeway starting with hypha server infra.

@bnmajor initialization, `await`'s, etc. needs tweaks?

Needs to be run with https://github.com/amun-ai/hypha/pull/432

oeway:
@thewtex https://github.com/amun-ai/hypha/pull/432 is merged now, docs here: https://ha.amun.ai/#/getting-started?id=serve-static-files

thewtex:
@oeway thank you!!

thewtex:
@oeway in the latest push, I am getting the errors in the command line console:

```
ERROR:asyncio:FuturePromise exception was never retrieved
future: <FuturePromise finished exception=Exception('Failed to get session store hosHnGWDuRUhAKD6djMqmo.PxPE2ynCb6vFYX9iXZBGEH')>
Exception: Failed to get session store hosHnGWDuRUhAKD6djMqmo.PxPE2ynCb6vFYX9iXZBGEH
```

why is this and what does it mean?

oeway:
It might because of the local proxy was removed when you call a certain function. I just tried your cli branch, the following worked:
```python
    ....
    await poll_for_services(server)
    print('services found')
    
    services = await server.list_services()
    service = [s for s in services if s["name"] == "itkwidgets_client"][0]
    itk_viewer = await server.get_service(service)
    image = itk.imread(data)
    await itk_viewer.setImage(np.array(image))
```
I can see the image get loaded in the opened browser tab.

I saw you have some polling functions in the cli.py, maybe a better way to implement it is to register a service in Python, e.g. a image zarr store provider service. We register it before we trigger the browser, when browser is loaded, the viewer use the workspace/token to get the zarr store from the service. Something like this:
```
await server.register_service({
      "name": "ItkwidgetsFileLoader",
      "id": "itkwidgets-file-loader",
      "description": "Load a file from the server",
      "config": {
          "visibility": "protected"
      },
      "load_file": load_file
})
```

In the itk-vtk-viewer, we can get the service and load the remote zarr store.

Another advantage is we can have multiple viewer window open at the same time. 

We can support collaborative visualisation too, since the server service can push viewer parameters to the viewer clients.

itkwidgets/standalone/index.html
```diff
@@ -0,0 +1,98 @@
...
+      const remoteService = await server.getService(`${remoteID}:itkwidgets-input-obj`);
+      const data = await remoteService.inputObject();
+      const viewer = await itkVtkViewer.createViewer(container, data.data);
+      await server.registerService({
+        name: "itk_vtk_viewer",
+        id: "itk-vtk-viewer",
+        description: "Retrieve the itk-vtk-viewer object",
+        config: {
+          visibility: "protected",
+          requireContext: false,
```
oeway, line 75:
```suggestion
          require_context: false,
```

itkwidgets/standalone/index.html
```diff
@@ -0,0 +1,98 @@
...
+          runInExecutor: false,
```
oeway, line 76:
```suggestion
          run_in_executor: false,
```

These config are in snake_case: https://github.com/imjoy-team/imjoy-rpc/blob/43c48a13840107f6a5c3964b8f0f062159964c83/javascript/src/hypha/rpc.js#L405-L410

bnmajor: Fixed, thank you!!

bnmajor:
@thewtex @oeway Thank you both for all of your help understanding what we actually need here!

This branch has been updated to register services for communication between the client and server rather than registering a plugin and seems to work well for now! There is basic arg support for some of the args the [createViewer](https://github.com/Kitware/itk-vtk-viewer/blob/master/src/createViewer.js#L29-L41) expects and can be expanded to add support for all of the other kwargs that `view` [currently accepts](https://github.com/InsightSoftwareConsortium/itkwidgets/blob/main/itkwidgets/viewer.py#L365-L470). It is still hard-coded to handle images using `itk.imread` and does not currently handle `point_set` files though. Before doing too much here though I thought it might be good to discuss best options for how to handle all the possible file inputs.

Additionally there is a convenience function now, `standalone_viewer`, which can be used to retrieve the `itk-vtk-viewer` object when a viewer has been created and served. Right now it just expects the full URL which is then parsed to use the workspace and token to connect to the server before using the service to get the viewer. The easiest way to use this is by starting `IPython` from the command line:

```
In [1]: from itkwidgets import standalone_viewer
In [2]: url = "http://127.0.0.1:37480/itkwidgets/index.html?workspace={...}&token={...}"
In [3]: viewer = await standalone_viewer(url)
In [4]: await viewer.setRotateEnabled(True)
```

This is achievable through the Python REPL as well but it is a bit more complicated:
```
>>> from itkwidgets import standalone_viewer
>>> import asyncio
>>> async def start_rotate(loop, url):
...     viewer = await standalone_viewer(url)
...     await viewer.setRotateEnabled(True)
>>> url = "http://127.0.0.1:37480/itkwidgets/index.html?workspace={...}&token={...}"
>>> loop = asyncio.new_event_loop()
>>> future = loop.create_task(start_rotate(loop, url))
>>> loop.run_until_complete(future)

```

oeway:
FYI: i recently added “startup function” support which might be useful for this cli. Not sure how well it will fit the cli case though: https://ha.amun.ai/#/?id=custom-initialization-and-service-integration-with-hypha-server

oeway:
Hi @bnmajor I added a synchronous wrapper for imjoy-rpc with hypha server: https://github.com/imjoy-team/imjoy-rpc/pull/546 (It works by running the event loop in a separate thread)
This might improve how the user use the itkwidgets cli since they won't need to use `async/await`. Let me know if it works for you.

bnmajor:
> FYI: i recently added “startup function” support which might be useful for this cli. Not sure how well it will fit the cli case though: https://ha.amun.ai/#/?id=custom-initialization-and-service-integration-with-hypha-server

@oeway I did look into using this approach, but it seems that the server object that is returned has no `register_codec` function - is this intentional?

oeway:
> @oeway I did look into using this approach, but it seems that the server object that is returned has no `register_codec` function - is this intentional?

Have you tried the latest version, it should work:
```python
>>> from imjoy_rpc.hypha import connect_to_server_sync
>>> server = connect_to_server_sync({"server_url": "https://ai.imjoy.io"})
>>> server.register_codec
<bound method RPC.register_codec of <imjoy_rpc.hypha.rpc.RPC object at 0x7fc248082520>>
```

oeway:
BTW, I just added webrtc support for imjoy-rpc: https://github.com/imjoy-team/imjoy-rpc/pull/547
This might be interesting for you too, it enables peer-to-peer connection between hypha clients, you can use our public server for signaling, the the rest will be p2p. We are setting up a STUN/TURN server for complicated firewall too. The plan is to support streaming microscope images and control it with hypha. It will be interesting for itkwidgets to enable remote rendering too!

bnmajor:
> > @oeway I did look into using this approach, but it seems that the server object that is returned has no `register_codec` function - is this intentional?
> 
> Have you tried the latest version, it should work:
> 
> ```python
> >>> from imjoy_rpc.hypha import connect_to_server_sync
> >>> server = connect_to_server_sync({"server_url": "https://ai.imjoy.io"})
> >>> server.register_codec
> <bound method RPC.register_codec of <imjoy_rpc.hypha.rpc.RPC object at 0x7fc248082520>>
> ```

Thanks @oeway! That is great to know. What I am actually referring to is the `server` object that is passed into the startup function:
```
with subprocess.Popen([
    sys.executable, "-m", "hypha.server",
    f"--host={SERVER_HOST}", f"--port={PORT}",
    "--static-mounts", f"/itkwidgets:{viewer_mount_dir}",
    "--startup-functions=itkwidgets.standalone_server:start_viewer"
], env=hypha_server_env)

...

async def start_viewer(server):
    register_itkwasm_imjoy_codecs_cli(server)

...

File "/home/local/KHQ/brianna.major/itkwidgets/itkwidgets/imjoy.py", line 53, in register_itkwasm_imjoy_codecs_cli
    server.register_codec({'name': 'itkwasm-image', 'type': itkwasm.Image, 'encoder': encode_itkwasm_image})
TypeError: 'NoneType' object is not callable
```

oeway:
> register_codec

@bnmajor Oh, I see, sorry for the confusion. I have fixed it now in `hypha>=0.15.27`: https://github.com/amun-ai/hypha/pull/472

bnmajor:
> > What I am actually referring to is the server object that is passed into the startup function
> 
> @bnmajor Oh, I see, sorry for the confusion. I have fixed it now in `hypha>=0.15.27`: [amun-ai/hypha#472](https://github.com/amun-ai/hypha/pull/472)

Thanks! Unfortunately I'm still seeing the same issue... I'm using hypha 0.15.27 and imjoy-rpc 0.5.31.

oeway:
> > @bnmajor Oh, I see, sorry for the confusion. I have fixed it now in `hypha>=0.15.27`: [amun-ai/hypha#472](https://github.com/amun-ai/hypha/pull/472)
> 
> Thanks! Unfortunately I'm still seeing the same issue... I'm using hypha 0.15.27 and imjoy-rpc 0.5.31.

Hi, sorry, missed [another place](https://github.com/amun-ai/hypha/pull/473/files#diff-f4bec7b32a68fc98f1493ec0e60a496d3780fb44115ec772d02add8e5b3f686dR890-R892) which requires change, now added as a [test case](https://github.com/amun-ai/hypha/pull/473/files#diff-8671705963f63b1364b9b55aa5fe9bc04fadf3de047049170697036f82421665R527-R531). 
It should work with hypha>=0.15.28 (available soon).

itkwidgets/standalone/config.py
```diff
@@ -0,0 +1,6 @@
+from pathlib import Path
+
+SERVER_PORT = 37480
```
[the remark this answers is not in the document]
bnmajor: Yup, fixed!

pyproject.toml
```diff
@@ -37,7 +37,7 @@ keywords = [
 requires-python = ">=3.7"
```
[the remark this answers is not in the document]
bnmajor: Done :+1:

itkwidgets/__init__.py
```diff
@@ -7,3 +15,7 @@
   "view",
   "compare_images",
 ]
+
+if ENVIRONMENT is not Env.JUPYTERLITE:
+  from .standalone_server import standalone_viewer
+  __all__.append("standalone_viewer")
```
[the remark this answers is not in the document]
bnmajor: Moved to `urllib3` for now

itkwidgets/standalone_server.py
```diff
@@ -0,0 +1,340 @@
...
+                if response.ok:
+                    break
+            except RequestException:
+                pass
+            timeout -= 0.1
+            time.sleep(0.1)
+
+        loop.create_task(start_viewer(server_url))
+        loop.run_forever()
+        time.sleep(30)
```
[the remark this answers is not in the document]
bnmajor: Done

itkwidgets/standalone_server.py
```diff
@@ -0,0 +1,340 @@
...
+    hypha_server_env = os.environ.copy()
+
+    server_url = f"http://{SERVER_HOST}:{SERVER_PORT}"
+    viewer_mount_dir = str(Path(VIEWER_HTML).parent)
+
+    with subprocess.Popen(
+        [sys.executable, "-m", "hypha.server", f"--host={SERVER_HOST}",
+         f"--port={SERVER_PORT}", "--static-mounts",
+         f"/itkwidgets:{viewer_mount_dir}"],
+        env=hypha_server_env,
```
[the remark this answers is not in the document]
bnmajor: Changed

bnmajor:
> Hi, sorry, missed [another place](https://github.com/amun-ai/hypha/pull/473/files#diff-f4bec7b32a68fc98f1493ec0e60a496d3780fb44115ec772d02add8e5b3f686dR890-R892) which requires change, now added as a [test case](https://github.com/amun-ai/hypha/pull/473/files#diff-8671705963f63b1364b9b55aa5fe9bc04fadf3de047049170697036f82421665R527-R531). It should work with hypha>=0.15.28 (available soon).

Thanks for all of the updates and fixes! Unfortunately in testing out the functionality a bit more I think that this may not be the move for our use case. We are using the services to trigger the next steps in the pipeline server-side, and this requires sharing some global state variables. Using the the startup function creates a new instance of the module so this global state is not available. If I am maybe missing something here please let me know though!

bnmajor:
@thewtex This PR currently:
- Is started with the cli entry point: `itkwidgets --image {...} {...optional args}`
- Can accept a `--verbose` flag to output stdout to the terminal again, otherwise output is surpressed
- Can accept a `--repl` flag which will automatically start an interactive REPL session
  - This session makes the `viewer` object available and this is explained in the banner created with the session
- The issue mentioned above regarding setting the label image seems to be the same issue described [before] (https://github.com/InsightSoftwareConsortium/itkwidgets/pull/589#issuecomment-1586218575). This issue happens at the same place (in [updateRenderedImage](https://github.com/Kitware/itk-vtk-viewer/blob/400d12a88f30b19eb2552c753676548976b5bf28/src/Rendering/VTKJS/Images/updateRenderedImage.js#L56-L152)) and only for [setImage](https://github.com/Kitware/itk-vtk-viewer/blob/master/src/createViewer.js#L692-L708) and [setLabelImage](https://github.com/Kitware/itk-vtk-viewer/blob/master/src/createViewer.js#L874-L897) where we queue the functions. Creating and calling a simple asynchronous function succeeds.

oeway:
> Thanks for all of the updates and fixes! Unfortunately in testing out the functionality a bit more I think that this may not be the move for our use case. We are using the services to trigger the next steps in the pipeline server-side, and this requires sharing some global state variables. Using the the startup function creates a new instance of the module so this global state is not available. If I am maybe missing something here please let me know though!

I see, then I think you are right, startup functions won't work in this case.

itkwidgets/standalone_server.py
```diff
@@ -0,0 +1,387 @@
...
+            viewer = view(itk_viewer=svc.viewer())
+            banner = f"""
+                Welcome to the itkwidgets command line tool! Press CTRL+D or
+                run `exit()` to terminate the REPL session. Use the `viewer`
+                object to manipulate the viewer.
+            """
+            exitmsg = "Exiting REPL. Press CTRL+C to teminate CLI tool."
+            code.interact(banner=banner, local={"viewer": viewer}, exitmsg=exitmsg)
+
+
```
thewtex, line 196: Can we also add:

```
if __name__ == "__main__":
    cli_entrypoint()
```

?

bnmajor: Sure! But should this be
```
if __name__ == "__main__":
  cli_entrypoint()
```
instead so that we still parse any flags?

bnmajor:
@thewtex The bug that we discussed where we cannot load a label image still exists but it is not obvious why... Thanks to @PaulHax we know that the exact issue is that `getItem` is failing for some reason ([here](https://github.com/Kitware/itk-vtk-viewer/blob/master/src/IO/ZarrStoreParser.js#L17)). What I know so far:

- If we pass the image and label data to the [client](https://github.com/InsightSoftwareConsortium/itkwidgets/pull/589/files#diff-7a4094cc4d3b062720abae240c3de24427a008da277ddc7823e76bd162ddf309R66-R67) via the [inputObject service function](https://github.com/InsightSoftwareConsortium/itkwidgets/pull/589/files#diff-221980ade0acb35265916827c4678d1ea5dafef99358abc1e6ad6dbc4a1b1f18R111-R123) and then we pass in all of the data *without* the label image we can call `await viewer.setLabelImage` right after `viewer` resolves and it works.
- If we try to call `viewer.set_label_image` (the wrapper for the client-side `viewer.setLabelImage` function) from the Python REPL session it fails at `getItem`.
- If we create a client-side service function that calls `viewer.setLabelImage` and pass it the image after it is loaded from the Python REPL session it fails at `getItem`.
- All approaches are taking the same path on the server-side to pass the same multiscale zarr store to the client.
- The only other difference I can see between the successful and failed approaches is that it appears that any time the client is in charge of requesting the data we have no issues, but any time that the server sends the data without a client request it fails. This makes me think maybe there is a codec issue? But if so shouldn't we see the same issue in the notebooks where we register the same codecs and pass the data from the server in the same manner?

itkwidgets/standalone_server.py
thewtex: yes :-)

thewtex:
@bnmajor great work debugging.

Another debugging session with @PaulHax , a few more interesting points:

- We confirmed that `getItem` is not returning on the JavaScript size, but only with the array chunks -- it succeeds for .zattrs and .zarray.
- We tried a few more approaches, and we confirmed your observation that it succeeds only when pulling from the JavaScript side. The call flow: `set_label_image` passes a remote zarr `store` on which the JS side will make remote procedure calls before completing do not succeed.

It appears that there is some issue with our syncified Python RPC that requires JS RPC calls before returning?

In any case, is there a workaround that we can use to behave like the initial input data objects, pulling client-side?

oeway:
> @bnmajor great work debugging.
> 
> Another debugging session with @PaulHax , a few more interesting points:
> 
> * We confirmed that `getItem` is not returning on the JavaScript size, but only with the array chunks -- it succeeds for .zattrs and .zarray.
> * We tried a few more approaches, and we confirmed your observation that it succeeds only when pulling from the JavaScript side. The call flow: `set_label_image` passes a remote zarr `store` on which the JS side will make remote procedure calls before completing do not succeed.
> 
> It appears that there is some issue with our syncified Python RPC that requires JS RPC calls before returning?
> 
> In any case, is there a workaround that we can use to behave like the initial input data objects, pulling client-side?

I think what you are observing is related to a dead lock in the syncified python RPC. It happens because we execute the syncified functions in the main thread. If we have a service function A registered, normally, it will respond to remote calls from the js side. However, if we initiate a call from the same main thread to call something on the js side, then the js side try to call A again, the thread gets locked because it will occupy the thread and function A will never get executed.

I need to check if I can solve it inside the rpc library. But for now, an easy way to solve this is to create two server connections and use one to register the internal services, and the other one for accessing from the Python REPL.

itkwidgets/standalone/index.html
```diff
@@ -77,6 +77,21 @@
...
         viewer: () => viewer
       });
+      await server.registerService({
+        name: "setLabelImage",
+        id: "set-label-image",
+        description: "Set the label image.",
+        config: {
+          visibility: "protected",
+          require_context: false,
+          run_in_executor: true,
```
thewtex, line 87: `false`?

oeway: This won't have any effect in js.

thewtex:
I added a commit that adds terminal output by default! :computer: :nerd_face: We can still open in a browser with the `-b` or `--browser` flag, but this enables quick terminal-based inspection of 3D data. It requires a terminal that supports the iterm2 inline image protocol such as:

- [wezterm](https://wezfurlong.org/wezterm/)
- VSCode's Terminal (with the latest version of VSCode)
- iTerm2

In VSCode:

[itkwidgets-vscode.webm](https://github.com/InsightSoftwareConsortium/itkwidgets/assets/25432/ab58ef99-a6bd-4ee6-a760-6aff6f1dea6f)

thewtex:
VSCode terminal + tmux. Requires `set-option -g allow-passthrough on` in *~/.tmux.conf*.

[itkwidgets-vscode-tmux.webm](https://github.com/InsightSoftwareConsortium/itkwidgets/assets/25432/f035bb58-eeaf-4790-bd6b-d3d36687df74)

thewtex:
wezterm on Linux:

[itkwidgets-wezterm-linux.webm](https://github.com/InsightSoftwareConsortium/itkwidgets/assets/25432/46540456-57f0-4d5e-b4bc-626eb4a8a46f)

thewtex:
Verified on Windows 10, wezterm, ssh'ing into an ARM mac:

image

and Windows ssh'ing into a headless Linux:

image

wezterm with Windows 11 WSL with itkwidgets running on Windows also works 🎉 🌮 

Install command will be:

```sh
pip install 'itkwidgets[cli]'
playwright install --with-deps chromium
```

thewtex (commented):
:100: :champagne: :sparkler: :sparkling_heart: 

WooT! wOOt!

As discussed with @PaulHax , for performance reasons, we are moving to having all the input datasets managed by a `DataManager` for the viewer that lives in a WebWorker. In that case, we will have a separate JS event loop in the web worker and separate connections to hypha -- so the workaround we have now will likely be a non-issue.

PR #591, opened 2023-01-04
Update to JupyterLite 0.1.0b17 and the `piplite_urls` config

jtpio:
With JupyterLite `0.1.0b17`, the `piplite_urls` configuration has changed and should now be put under the `PipliteAddon` field: https://github.com/jupyterlite/jupyterlite/releases/tag/v0.1.0b17

This change also pins to `jupyterlite[all]==0.1.0b17` for now to avoid catching potential breaking changes during the beta cycle.


thewtex (approved):
:1234: :+1: @jtpio thanks!

PR #599, opened 2023-01-17
DOC: Thin plate spline example updates

thewtex:
- Add missing package dependencies
 - Fix kwarg: point_sets -> point_set
 - Use web3.storage for data storage -- the hash does not appear to be stable


thewtex:
Merging to resolve CI greenness on main -- data hash has changed.

PR #600, opened 2023-01-17
WIP: DOC: Reload examples page if itkwidgets[lab] installed later

thewtex:
If the imjoy jupyterlab extension is installed after JupyterLab has been loaded, the page needs to be reloaded to load up the added JavaScript.


bnmajor:
@thewtex In my testing this works exactly as expected for locally hosted jupyter instances but not in SageMaker... It prompts to refresh as expected but refreshing does not make the extension available and the viewer is still not output. Trying to pin down the why.

PR #601, opened 2023-01-17
COMP: Init with fiji imagej

thewtex:
To address:

```
Failed to bootstrap the artifact.

Possible solutions:
* Double check the endpoint for correctness (https://search.maven.org/).
* Add needed repositories to ~/.jgorc [repositories] block (see README).
* Try with an explicit version number (release metadata might be wrong).

Full Maven error output:
Error: R] Some problems were encountered while processing the POMs:
Error: ] Non-resolvable import POM: Could not transfer artifact net.imagej:imagej:pom:2.5.0 from/to scijava.public (https://maven.scijava.org/content/groups/public): transfer failed for https://maven.scijava.org/content/groups/public/net/imagej/imagej/2.5.0/imagej-2.5.0.pom @ line 8, column 29
Error: ] 'dependencies.dependency.version' for net.imagej:imagej-legacy:jar is missing. @ line 10, column 252
Error: ] 'dependencies.dependency.version' for org.scijava:scijava-config:jar is missing. @ line 10, column 462
Error: ] The build could not read 1 project -> [Help 1]
Error: ]
Error: ]   The project net.imagej-BOOTSTRAPPER:imagej-BOOTSTRAPPER:0 (/home/runner/.jgo/net.imagej/imagej/2.5.0/25280ba63ac2b82ba82d0e0126a26046929ed9f235535b7c20636efa1a51762f/pom.xml) has 3 errors
Error: ]     Non-resolvable import POM: Could not transfer artifact net.imagej:imagej:pom:2.5.0 from/to scijava.public (https://maven.scijava.org/content/groups/public): transfer failed for https://maven.scijava.org/content/groups/public/net/imagej/imagej/2.5.0/imagej-2.5.0.pom @ line 8, column 29: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target -> [Help 2]
Error: ]     'dependencies.dependency.version' for net.imagej:imagej-legacy:jar is missing. @ line 10, column 252
Error: ]     'dependencies.dependency.version' for org.scijava:scijava-config:jar is missing. @ line 10, column 462
Error: ]
Error: ] To see the full stack trace of the errors, re-run Maven with the -e switch.
Error: ] Re-run Maven using the -X switch to enable full debug logging.
Error: ]
Error: ] For more information about the errors and possible solutions, please read the following articles:
Error: ] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/ProjectBuildingException
Error: ] [Help 2] http://cwiki.apache.org/confluence/display/MAVEN/UnresolvableModelException
```

in CI.

### DIFF
--- a/.github/workflows/notebook-test.yml
+++ b/.github/workflows/notebook-test.yml
@@ -29,7 +29,7 @@ jobs:
           python3 -m pip install --upgrade pip
           python3 -m pip install -e ".[test,all]"
           python3 -m pip install pyimagej
-          python3 -c "import imagej; ij = imagej.init('2.5.0'); print(ij.getVersion())"
+          python3 -c "import imagej; ij = imagej.init('sc.fiji:fiji:2.5.0'); print(ij.getVersion())"
           python3 -m pip install "itk>=5.3.0"
 
       - name: Test notebooks


thewtex:
My naive attempt to change the source following pyimagej did not work.

@ctreuden any ideas?

ctrueden:
I think the problem is:

> PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target

But the question is why. I've seen this error before when the version of Java is older than 1.8.0_101. But from the build log:

> Downloading Java 8.0.352+8 (Zulu) from https://cdn.azul.com/zulu/bin/zulu8.66.0.15-ca-jdk8.0.352-linux_x64.tar.gz ...

I'm downloading that same version now, and testing locally.

thewtex:
@ctrueden thanks for taking a look!

Do you recommend updating the version here:

https://github.com/InsightSoftwareConsortium/itkwidgets/blob/0e42c5d4a56b456d9a85287f633e1d3bd3ce59d0/.github/workflows/notebook-test.yml#L24-L25

?

ctrueden:
Nope, I was wrong, it's not Zulu-version-specific, it's actually intermittent. 😮 So it just *happened* to fail in my first 8u352 and 8u322 tests, and work in my first 8u252 test, but I've not gotten it to both succeed and fail for all versions by repeating the build. Still investigating. For now I suggest not changing anything, since I don't think it will help the situation.

ctrueden:
@thewtex I believe the problem is fixed now. It seemed to be an issue with there being multiple Apache processes, some of which were stuck serving the previous expired SSL cert, creating the observed nondeterminism. After doing a full restart of Apache, I'm not able to replicate the issue on any system anymore. So try restarting the build and see if it works now! (As an aside: it shouldn't be necessary to use `sc.fiji:fiji` in the endpoint unless you really do want to have access to Fiji-specific things in the environment.)

thewtex:
@ctrueden Curtis, that did it, thank you!! :tada: :taco: :pray: 

In this case, we will keep the current distribution to limit the download for folks wanting to try pyimagej quickly, so closing.

PR #602, opened 2023-01-19
ENH: Set max concurrency to server logical processors times two

bnmajor:


bnmajor:
@thewtex This update works but does not seem to resolve any of the issues around SageMaker Studio's silent failing

thewtex (commented):
@bnmajor thanks for testing! Nothing? Even if the `* 2` is removed?

Overall, an improvement nonetheless.

Could the limit please checked and excluded for the JupyterLite case if  `cpu_count` is not correctly reported?

bnmajor:
@thewtex This has been tested with Colab and JupyterLite now with no issues.

Side note this also worked with Studio once I deleted my existing user and created a new one (worked on every try). Not sure that this is the solution (1.0a22 started working after I tested this as well) but at least it certainly was not creating new issues.

thewtex:
@bnmajor thanks!

Please rebase/merge/release.

> Side note this also worked with Studio once I deleted my existing user and created a new one (worked on every try). Not sure that this is the solution (1.0a22 started working after I tested this as well) but at least it certainly was not creating new issues.

Sounds positive! :crossed_fingers: :1st_place_medal: 

CC @aylward

thewtex (approved):

PR #607, opened 2023-01-24
BUG: Fix keyword param point_set -> point_sets

bnmajor:


thewtex (commented):
I think this should be `point_set` -- so adjustments may need to be made to the `view` code?

In 0.X, we could pass a list of point sets (and geometries), however in 1.X, I would like to support a single point set and geometry for `view` because it avoid ambiguities with NumPy-derived data types and lists and it provide a more reasonable way to specify parameters -- they can just be passed as kwargs to `view`. To add more data the scene, `viewer = view(point_set1); viewer.set_point_set(point_set2)` Does that make sense?

bnmajor:
@thewtex Makes sense! I'll just update this PR to make those fixes instead

bnmajor:
@thewtex Any preference on the default behavior for detecting the input type for [numpy arrays](https://github.com/InsightSoftwareConsortium/itkwidgets/blob/main/itkwidgets/integrations/__init__.py#L166-L170), [zarr groups](https://github.com/InsightSoftwareConsortium/itkwidgets/blob/main/itkwidgets/integrations/__init__.py#L171-L175), [dask arrays](https://github.com/InsightSoftwareConsortium/itkwidgets/blob/main/itkwidgets/integrations/__init__.py#L192-L196), [PyTorch tensors](https://github.com/InsightSoftwareConsortium/itkwidgets/blob/main/itkwidgets/integrations/__init__.py#L199-L203), or [xarray DataArrays or Datasets](https://github.com/InsightSoftwareConsortium/itkwidgets/blob/main/itkwidgets/integrations/__init__.py#L206-L215)? In all of these cases the default is to treat the data as an image if not explicitly passed in with the `point_set` kwarg. So for your example `viewer = view(point_set1)` this would (for most data types) be treated as an image without `point_set` keyword.

Also right now the behavior is to [overwrite](https://github.com/Kitware/itk-vtk-viewer/blob/master/src/createViewer.js#L386-L388) the existing point set rather than update the list. I could either update `setPointSets` to append the new data, or maybe it would make more sense to add a new function like `updatePointSets` or `appendPointSet`?

thewtex:
@bnmajor thanks!

> @thewtex Any preference on the default behavior for detecting the input type for numpy arrays, zarr groups, dask arrays, PyTorch tensors, or xarray DataArrays or Datasets? In all of these cases the default is to treat the data as an image if not explicitly passed in with the point_set kwarg. So for your example viewer = view(point_set1) this would (for most data types) be treated as an image without point_set keyword.

In detecting the type without an explicit kwarg, can we have `Nx3` (x,y,z) or `Nx2` (x,y) imply point set, and image otherwise?

> Also right now the behavior in the itk-vtk-viewer api is to overwrite the existing point set rather than update the list. I could either update setPointSets to append the new data, or maybe it would make more sense to add a new function like updatePointSets or appendPointSet?

Thinking about transitions, `addPointSet` would be ideal.

itkwidgets/_type_aliases.py
```diff
@@ -15,21 +15,21 @@
 Style = Dict[str, str]
 
 Image = Union[np.ndarray, itkwasm.Image, zarr.Group]
-Point_Sets = Union[np.ndarray, itkwasm.PointSet, zarr.Group]
+Point_Set = Union[np.ndarray, itkwasm.PointSet, zarr.Group]
```
thewtex, line 18: unrelated to your changes here, but can we please fix regardless: `Point_Set` -> `PointSet` :pray:

bnmajor:
> In detecting the type without an explicit kwarg, can we have `Nx3` (x,y,z) or `Nx2` (x,y) imply point set, and image otherwise?

Sure thing! Updated with those changes now :+1: 

> Thinking about transitions, `addPointSet` would be ideal.

Awesome. This branch now relies on Kitware/itk-vtk-viewer#643. Once that is in the itk-vtk-viewer version will need to be bumped before this branch is ready.

itkwidgets/_type_aliases.py
bnmajor: Fixed!

thewtex (approved):
:1st_place_medal: :100:

PR #609, opened 2023-01-24
Environment yaml

bnmajor:


bnmajor:
@thewtex This doesn't include nodejs at the moment even though that is required for SageMaker for now... Should I add that in and just come back to remove it once we've simplified the SageMaker installation process?

thewtex:
> @thewtex This doesn't include nodejs at the moment even though that is required for SageMaker for now... Should I add that in and just come back to remove it once we've simplified the SageMaker installation process?

@bnmajor Yes, good plan. :+1:

bnmajor:
> Yes, good plan. +1

@thewtex Updated :+1:

thewtex (approved):

PR #618, opened 2023-02-23
TST: Disable testing for Python 3.11

bnmajor:
Temporary fix to avoid issues with `wasmer-python` since they have no 3.11 wheels available.


bnmajor:
Issue exists for this: wasmerio/wasmer-python#696

When this issue is resolved or `itk-wasm` uses an alternatice to `wasmer-python` we should re-enable testing for 3.11.

thewtex (approved):
@bnmajor thanks!

PR #634, opened 2023-03-14
DOC: Add version query param to JupyterLite URL

thewtex:
This attempts to add more cache refreshing so the latest version of the notebook / JupyterLite deployment is retrieved.

However, testing locally, it did provide an updated notebook for me with Chrome, npm's http-server.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -10,7 +10,7 @@ powered by [itk-wasm](https://wasm.itk.org) and [ImJoy](https://imjoy.io/).
 *Welcome to The Future* 🔬🚀🛸!
 ```
 
-<a href="./_static/retro/notebooks/?path=Hello3DWorld.ipynb">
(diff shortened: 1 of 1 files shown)


oeway (commented):
Unfortunately, it didn't work for me, I think the caching is not in the browser, but it's in the jupyterlite storage layer (implemented via localForage I guess: https://jupyterlite.readthedocs.io/en/latest/howto/configure/storage.html).

With the Imjoy extension, we support `load=raw github/gist url` and `open=1` but it seems failing for your file. The one worked is: https://jupyter.imjoy.io/lab/index.html?load=https://gist.github.com/oeway/391b4352ea57b5682366ce3dc2fa9174&open=1

This we make sure it loaded directly from the url, not the local storage.

I think the bug is coming from the file browser in jupyterlab, I just filed an issue here: https://github.com/jupyterlab/jupyterlab/issues/14194

PR #641, opened 2023-04-04
BLD: Bump itk-vtk-viewer 14.35.0 -> 14.35.1

bnmajor:
This limits the step value for the window/level interactor to 10 (values >= 100 reduce reactivity to near nothing).


thewtex (approved):
:rocket:

PR #644, opened 2023-04-17
BLD: Bump itk-vtk-viewer 14.35.0 -> 14.36.0

bnmajor:
Adds the following changes from itk-vtk-viewer:
- Handle TAKE_SCREENSHOT event: This should hopefully prevent the occasional "No object bound to API" error we were seeing in Colab notebooks
- do not clamp the color range


thewtex (approved):
:rocket:

PR #647, opened 2023-04-28
FIX: Do not ignore keyword args passed to viewer

bnmajor:
Keyword args were being lost. Make sure they're not.


thewtex (approved):
:1234:

PR #649, opened 2023-05-05
COMP: Add .readthedocs.yml

thewtex:
Explicitly use newer Python.


thewtex:
Attempt to address build failures observed in #648

bnmajor (approved):
Awesome, thank you!

PR #656, opened 2023-05-19
FIX: bump itk-vtk-viewer to 14.41.3

PaulHax:
Incorporates fixes for negative direction cosines and different number of scales in fixed and moving images.


bnmajor (approved):
Works well for me!

thewtex (commented):
:100:

PR #657, opened 2023-05-19
DOC: Add a notebook showing off example IDC data

bnmajor:
This adds a new notebook that allows users to simply download and visualize some of the example data from the IDC segmentation primer notebook, as well as additional dataset provided by @aylward (there was a specific request for MR data which was not available in their example). Each section also includes a link to simply open the itk-vtk-viewer with the data without having to download the data.

A first pass at the notebook, open to any suggestions/changes (including naming and location in the examples directory).


thewtex (approved):
:sparkler: :rocket: 

@bnmajor thank you!!

PR #664, opened 2023-05-24
fetch DICOM CT and SEG from IDC

fedorov:
I marked this as draft since:
1. I did not see segmentation rendering - not sure if I need to do anything special.
2. the new CT series that I selected does not have the corresponding RTSTRUCT - I do not know if you wanted that to be in the notebook, or it was optional.


fedorov:
@bnmajor I think you should have been able to push to the branch in this PR, since I have "Allow edits by maintainers" checked. Maybe you can try this next time instead of creating a new PR (should be less work) ;-)

This PR has been superseded by #665.

PR #669, opened 2023-06-01
ENH: Add bounding box feature

### LINKED ISSUE #661: Color range is not correct in with compare_images() in MultiScale GYBT notebook

### LINKED ISSUE #666: compare_image argument method default should be cyan-magenta
Seems to default to `disabled`

bnmajor:
Allows user to display a bounding box around each of the image layers.

fixes #661
fixes #666


PaulHax (approved):
🍱

PR #701, opened 2023-12-02
ENH: register callback for saveRoi viewer event

PaulHax:
Depends on https://github.com/Kitware/itk-vtk-viewer/pull/723

This is the `viewer_config.py` file I'm using to test:
```
ITK_VIEWER_SRC = (
    "http://localhost:8082"
)
PYDATA_SPHINX_HREF = "http://localhost:3000/dist/bootstrapUIMachineOptions.js.es.js"
MUI_HREF = "https://cdn.jsdelivr.net/npm/itk-viewer-material-ui@0.3.0/dist/materialUIMachineOptions.js.es.js"
```

### DIFF
--- a/itkwidgets/viewer.py
+++ b/itkwidgets/viewer.py
@@ -82,6 +82,9 @@ class ViewerRPC:
                 itk_viewer.registerEventListener(
                     'screenshotTaken', self.update_screenshot
                 )
+                itk_viewer.registerEventListener(
+                    'saveRoi', self.save_roi
+                )
                 # Once the viewer has been created any queued requests can be run
                 CellWatcher().update_viewer_status(self.parent)
                 asyncio.get_running_loop().call_soon_threadsafe(self.viewer_event.set)
@@ -124,6 +127,9 @@ class ViewerRPC:
                 </script>
             ''')
         self.img.display(html)
+    
+    def save_roi(self, roi):
+        api.log(f'save roi {roi}')
 
     def set_event(self, event_data):
         # Once the data has been set the deferred queue requests can be run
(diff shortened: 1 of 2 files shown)


itkwidgets/viewer.py
```diff
@@ -124,6 +127,9 @@ def update_screenshot(self, base64_image):
                 </script>
             ''')
         self.img.display(html)
+    
+    def save_roi(self, roi):
+        print('save roi', roi)
```
PaulHax, line 132: I'm not seeing this actualy print, so missing something...

bnmajor: Print statements do not bubble up correctly from the ViewerRPC class. You can use `api.log(f'save roi {roi}')` instead and the message will be output in the console log

PaulHax: Ah, thank you.  This is the output: 
```
Plugin 5b4de75c-9cb8-490b-8c32-28c9ae68f0c4: save roi {'name': 'Image', 'layerName': 'Image', 'croppingPlanes': [{'origin': [-0.5, -0.5, -0.5], 'normal': [1, 0, 0]}, {'origin': [127.5, 127.5, 255.5], 'normal': [-1, 0, 0]}, {'origin': [-0.5, -0.5, -0.5], 'normal': [0, 1, 0]}, {'origin': [127.5, 127.5, 255.5], 'normal': [0, -1, 0]}, {'origin': [-0.5, -0.5, -0.5], 'normal': [0, 0, 1]}, {'origin': [127.5, 127.5, 255.5], 'normal': [0, 0, -1]}]}
```

PR #705, opened 2023-12-08
ENH: Support passing zarr stores in directly

bnmajor:


itkwidgets/integrations/__init__.py
```diff
@@ -59,6 +59,9 @@ def _get_viewer_image(image, label=False):
         to_ngff_zarr(store, image, chunk_store=chunk_store)
         return store
 
+    if isinstance(image, zarr.storage.Store):
```
thewtex, line 62: To support more cases, can we use `zarr.storage.BaseStore`?

thewtex (approved):
:sparkler:

PR #709, opened 2023-12-13
BUG: Select roi on itk image

thewtex:
Add a section to the Select ROI notebook on how to apply it to an itk.Image. Also fix a few bugs discovered in the process:

1. For slices, it should be index + 1 for the stop value
2. Ensure the slice indexes are integers.


bnmajor (approved):
Thank you!

thewtex:
Added, thanks for the pointer!

PR #716, opened 2023-12-19
Cell watcher docstrings

bnmajor:


itkwidgets/cell_watcher.py
```diff
@@ -1,69 +1,96 @@
...
-                names.append(name)
-        return names
+    def data(self) -> Dict[str, Dict[str, bool]]:
+        """Get the underlying data dict containg all viewer data
 
-    @property
-    def not_named(self):
-        # Return a list of names of viewers that have not been named yet
-        return any([k for k, v in self.data.items() if v['name'] is None])
+        :return: The data object that contains all created Viewer information.
```
thewtex, line 26: can we describe here what the `str`, `str`, `bool`'s are?

thewtex (changes requested):
Wow, awesome!

One minor request inline.

itkwidgets/cell_watcher.py
bnmajor: Sure thing!

thewtex (approved):
:100:

PR #724, opened 2023-12-20
ENH: Add vmin and vmax to viewer

### LINKED ISSUE #723: vmin and vmax viewer properties set color range
I don't know how to wrap get/set of color range with matplotlib like vmin and vmax properties.  How to do synchronous wrapping of async Viewer methods?

Goal is something like adding this to Viewer class
```
    @property
    def vmin(self):
        range = self.get_image_color_range()
        return range[0]
    @vmin.setter
    def vmin(self, vmin):
        range = self.get_image_color_range()
        range[0] = vmin
        self.set_image_color_range(range)
```

PaulHax:
Closes #723

Depends on:
https://github.com/Kitware/itk-vtk-viewer/pull/730#pullrequestreview-1793586609

### DIFF
--- a/itkwidgets/_initialization_params.py
+++ b/itkwidgets/_initialization_params.py
@@ -15,6 +15,8 @@ def init_params_dict(itk_viewer):
         'blend_mode': itk_viewer.setImageBlendMode,
         'cmap': itk_viewer.setImageColorMap,
         'color_range': itk_viewer.setImageColorRange,
+        'vmin': itk_viewer.setImageColorRangeMin,
+        'vmax': itk_viewer.setImageColorRangeMax,
         'color_bounds': itk_viewer.setImageColorRangeBounds,
         'component_visible': itk_viewer.setImageComponentVisibility,
         'gradient_opacity': itk_viewer.setImageGradientOpacity,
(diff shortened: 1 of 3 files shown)


thewtex (commented):
Nice!

@bnmajor if / when we have an inherited ipywidget viewer class, they will be able to override these properties with the traitlets, correct?

@PaulHax could you please also add `vmin`, `vmax` arguments to `view`?

bnmajor:
> @bnmajor if / when we have an inherited ipywidget viewer class, they will be able to override these properties with the traitlets, correct?

Yup! Essentially you would just swap out `property` for `observe` and `val.setter` for `default("val")`. For example:

```
class Foo:
    ....
     @property
     def bar(self):
             return self.get_value()

     @bar.setter
     def bar(self, val):
             self.set_value(val)
```

should become

```
class Foo(HasTraits):
    bar = Int()

     @observe
     def bar(self):
             return self.get_value()

     @default("bar")
     def bar(self, val):
             self.set_value(val)
```

thewtex (approved):
:+1: please bump the itk-vtk-viewer version before merge

thewtex:
tagging

PR #725, opened 2023-12-23
ENH: update transfer function editor color range features

PaulHax:
* Add intensity data value tooltip to control points
* Move color range control points lower
* Add middle color range control point


thewtex (approved):
:champagne:

PR #726, opened 2023-12-29
Do not use CellWatcher outside of notebook env

bnmajor:


itkwidgets/viewer.py
```diff
@@ -40,7 +40,8 @@
 
 _viewer_count = 1
 _codecs_registered = False
```
thewtex, line 42: Can we have something like:

```
_cell_watcher = None
if ENVIRONMENT is not Env.HYPHA:
    _cell_watcher = CellWatcher() # Instantiate the singleton class right away

[...]
# CellWatcher().add_viewer(self.parent) to
_cell_watcher and _cell_watcher.add_viewer(self.parent)
```

thewtex (changes requested):
@bnmajor thanks!

Minor refactor requested inline.

itkwidgets/viewer.py
bnmajor: Sure thing!

thewtex (approved):
:100:

PR #727, opened 2023-12-31
Docstrings

bnmajor:
Add docstrings to the `ViewerRPC` and `Viewer` classes

### DIFF
--- a/itkwidgets/_type_aliases.py
+++ b/itkwidgets/_type_aliases.py
@@ -16,6 +16,8 @@ Style = Dict[str, str]
 
 Image = Union[np.ndarray, itkwasm.Image, zarr.Group]
 PointSet = Union[np.ndarray, itkwasm.PointSet, zarr.Group]
+CroppingPlanes = List[Dict[str, List[Union[float, int]]]]
+
 if HAVE_ITK:
     import itk
     Image = Union[Image, itk.Image]
--- a/itkwidgets/_type_aliases.py
+++ b/itkwidgets/_type_aliases.py
@@ -7,7 +7,7 @@ from .integrations.itk import HAVE_ITK
 from .integrations.pytorch import HAVE_TORCH
 from .integrations.vtk import HAVE_VTK
 from .integrations.xarray import HAVE_XARRAY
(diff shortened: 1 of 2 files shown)


itkwidgets/_type_aliases.py
```diff
@@ -16,6 +16,8 @@
 
 Image = Union[np.ndarray, itkwasm.Image, zarr.Group]
 PointSet = Union[np.ndarray, itkwasm.PointSet, zarr.Group]
+CroppingPlanes = List[Dict[str, List[Union[float, int]]]]
```
thewtex, line 19: This is a nice addition.

We could make the types more self-documenting by using string `Literal`'s like in:

https://github.com/spatial-image/spatial-image/blob/main/spatial_image.py#L23

itkwidgets/viewer.py
```diff
@@ -315,118 +456,313 @@ async def get_image(self, name: str = 'Image') -> NgffImage:
...
+
+        :return: The future for the coroutine, to be updated with the current
+        blend mode.
+        :rtype:  asyncio.Future | str
+        """
         return await self.viewer_rpc.itk_viewer.getImageBlendMode()
 
     @fetch_value
-    def set_image_color_map(self, colorMap: str):
+    def set_image_color_map(self, colorMap: str) -> None:
```
thewtex, line 479: We should use snake_case for the arguments to keep the Python interfaces idiomatic Python when possible, i.e. colorMap -> color_map

thewtex (changes requested):
Absolutely fantastic! :1st_place_medal: :book: 

A few minor requests inline.

itkwidgets/_type_aliases.py
bnmajor: Awesome!

itkwidgets/viewer.py
bnmajor: Yes, makes sense, thanks!

thewtex (approved):
Awesome, thank you!!

Can we please rebase on `main`?

thewtex (approved):
:book:

PR #731, opened 2024-01-04
Jupyterlite tracebacks

thewtex:


thewtex:
We will need to update the JupyterLite version in the docs build after this is merged and released to fix the docs JupyterLite.

PR #741, opened 2024-04-14
ENH: Remove imjoy-jupyterlab-extension pin for JupyterLab>=4 support

thewtex:
Addressed with: https://github.com/imjoy-team/imjoy-jupyterlab-extension/pull/7

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,13 +61,13 @@ Source = "https://github.com/InsightSoftwareConsortium/itkwidgets"
 
 [project.optional-dependencies]
 all = [
-    "imjoy-jupyterlab-extension ==0.1.13",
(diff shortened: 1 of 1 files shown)


thewtex:
Thank you @oeway !

oeway:
Looks good! However, I am not sure it will work for the jupyterlab<4, is this backward compatible?

thewtex:
>  I am not sure it will work for the jupyterlab<4, is this backward compatible?

Good point. Probably not. And the JupyterLab docs recommend publishing a major version of the extension to go along with their breaking changes. However, since we are still alpha, and more breaking changes should be anticipated, I think we should still publish as 1.0aX.

PR #742, opened 2024-04-16
itk.image() to itk.output() in _type_aliases.py

GingerSpacetail:
I'm not sure yet, as the whole itk module is currently not working for me... But meanwhile you might want to check if this line should be updated to match the fact that itk.image() is deprecated, and we should use itk.output() instead

https://github.com/InsightSoftwareConsortium/ITK/blob/2a1265d75f96ff7a072791629ff23b8c42123b3d/Wrapping/Generators/Python/itk/support/extras.py#L137

### DIFF
--- a/itkwidgets/_type_aliases.py
+++ b/itkwidgets/_type_aliases.py
@@ -20,7 +20,7 @@ CroppingPlanes = {Literal['origin']: List[float], Literal['normal']: List[int]}
 
 if HAVE_ITK:
     import itk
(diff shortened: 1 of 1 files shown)


thewtex:
What issue are you seeing?

Note that `itk.Image` identifies a type, `itk.image` is a function to return the image instance that is either an `itk.Image` instance directly or the instance that would be produced by an `itk` filter.

Issue #743, opened 2024-04-17
Kernel hangs after itkwidgets visualization in jupyter notebook


KumoLiu:
- kernal hangs after view function in the notebook.
1. docker pull nvcr.io/nvidia/pytorch:24.03-py3
2. run the [example](https://github.com/InsightSoftwareConsortium/itkwidgets/blob/v1.0a50/examples/integrations/MONAI/transform_visualization.ipynb), kernal hangs after view called.

- After install `itkwidgets[notebook]>=1.0a49`, look like there will be a requirement compatibility issues for pandas, not sure whether it's related the issue.
```
Collecting pandas>=2.0.0 (from dask-image->ngff-zarr[dask-image]>=0.4.3->itkwidgets>=1.0a49->itkwidgets[notebook]>=1.0a49)
  Downloading pandas-2.2.2-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (19 kB)
```
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
cudf 24.2.0 requires pandas<1.6.0dev0,>=1.3, but you have pandas 2.2.2 which is incompatible.
dask-cuda 24.2.0 requires pandas<1.6.0.dev0,>=1.3, but you have pandas 2.2.2 which is incompatible.
dask-cudf 24.2.0 requires pandas<1.6.0dev0,>=1.3, but you have pandas 2.2.2 which is incompatible.
```
- After install the itkwidgets and exiting the Jupyter notebook, trying to launch the notebook again leads to a configuration error. It appears that the installation of itkwidgets is causing a disruption in the configuration settings.
<img width="1490" alt="Screen Shot 2024-04-17 at 15 51 33" src="https://github.com/Project-MONAI/tutorials/assets/55491388/c4d909d4-f6de-4ff5-8e3c-66b9fd4c94ff">

thewtex:
Addressed with #745

PR #744, opened 2024-04-17
COMP: Only depend on dask-image in pyodide

thewtex:
Avoid conflicts with pandas version (Issue #743) and keep light.


aylward (approved):
Fewer dependencies is always good :)

bnmajor (approved):

PR #748, opened 2024-04-18
ENH: Consolidate JUPYTERLAB and JUPYTER_NOTEBOOK Env's into JUPYTER

thewtex:
With Jupyter Lab 4 / Jupyter Notebook 7, we cannot detect the difference
and there is not a difference in how we deal with them.


bnmajor (approved):
:tada:

PR #755, opened 2024-06-21
feat: add set_image_piecewise_function_points

PaulHax:
remove set_image_piecewise_function_gaussians

Depends on
https://github.com/Kitware/itk-vtk-viewer/pull/751


itkwidgets/_type_aliases.py
```diff
@@ -7,10 +7,9 @@
 from .integrations.pytorch import HAVE_TORCH
 from .integrations.vtk import HAVE_VTK
 from .integrations.xarray import HAVE_XARRAY
-from typing import Dict, List, Literal, Union
+from typing import Dict, List, Literal, Union, Tuple
 
-Gaussian_Curve = Dict[str, float]
-Gaussians = Dict[str, List[Gaussian_Curve]]
+Points2d = List[Tuple[float, float]]
```
thewtex, line 12: This could be

```
Sequence[Sequence[float, float]]
```

? Please check if variants work by passing them to the `set_*` method. If they do not work, then it may be worthwhile type casting with `list()` or `tuple()` to provide a more convenient interface for the caller.

In the documentation example there is:

```
[[.2, .1], [.8, .9]]
```

which is a List[List[float, float]].

thewtex (changes requested):
@PaulHax awesome, thank you!

itkwidgets/_type_aliases.py
PaulHax: Ah, yes.  Updated type to `Sequence[Sequence[float]]`.  Sequence can't do `Sequence[float, float]`.  Passing tuple points or list points works.  Passing 3D point bugs out TF editor =/

itkwidgets/_initialization_params.py
```diff
@@ -22,7 +22,7 @@ def init_params_dict(itk_viewer):
         'gradient_opacity': itk_viewer.setImageGradientOpacity,
         'gradient_opacity_scale': itk_viewer.setImageGradientOpacityScale,
         'interpolation': itk_viewer.setImageInterpolationEnabled,
-        'gaussians': itk_viewer.setImagePiecewiseFunctionGaussians,
+        'piecewise_function_points': itk_viewer.setImagePiecewiseFunctionPoints,
```
thewtex, line 25: How about `tf_points` or something else shorter here?

thewtex: Or the more specific `transfer_function_points`?

PaulHax: Changed to `transfer_function_points`

thewtex: @PaulHax thank you! How about `transfer_function` for brevity? I do not think `_points` is adding enough information to warrant the length for interactive use. They are always points and this is implicit in the specification.

PaulHax: Better and better.  Changed.

thewtex (approved):
:100:

PR #764, opened 2024-10-21
Add simple ITK-Wasm DICOM SEG example

PaulHax:
supersedes #759


thewtex:
@PaulHax this is great!

Is this ready-to-go?

Rebase on `main` for CI?

thewtex:
`pydicom_seg` appears to be broken in newer Python -- I think we should use `itkwasm-dicom` instead.

thewtex:
Test failure:

```
FAILED examples/integrations/itkwasm/DICOM_SEG.ipynb:: - A cell timed out while it was being executed, after 3000 seconds.
```

PaulHax:
> pydicom_seg appears to be broken in newer Python / PyDICOM -- I think we should use itkwasm-dicom instead.

Yes, #757 does that.

PR #768, opened 2025-01-07
BUG: Move from on.fleek.co to dweb.link gateway

thewtex:
Fleek made a major transition to new infrastructure, their fleek.co
gateway is no longer available.

Move to the Protocol Labs public dweb.link gateway.

These CID's were also pinned on Pinata.


PaulHax (approved):
Works good

PR #774, opened 2025-03-13
DOC: Remove Colab example links

thewtex:
We will add these back after replacing the current implementation with a
more robust approach.


oeway:
FYI: It's a bit unfortunate that the colab is broken now, I won't have bandwidth to fix that right now, so let's disable it for now.

thewtex:
@oeway thanks :heart:

